PR #48, opened 2023-03-28
fix `setup.py`

Borda:
fix missing arguments anduse Utils to validate workflows and is the package can be installed


setup.py
```diff
@@ -13,4 +21,6 @@
         "bitsandbytes",
     ],
     packages=find_packages(),
+    long_description=readme,
```
williamFalcon, line 24: negative
@lantiga

williamFalcon: don't bloat this codebase please...
let's keep all this engineering stuff out (@lantiga )

Borda: ok, closing it; just be aware that the setup is not valid :)
also, let's state in the contribution section that we do not expect any code health contributions...

lantiga (approved):
Thanks for fixing setup.py, approved

PR #59, opened 2023-03-30
WIP: Alpaca LoRA finetuning (2/n)

awaelchli:


lit_llama/lora.py
```diff
@@ -35,7 +37,7 @@ def __init__(
         self.merge_weights = merge_weights
 
 
-class MergedLinear(nn.Linear, LoRALayer):
+class MergedLinear(bnb.nn.Linear8bitLt, LoRALayer):
```
awaelchli, line 40: See: https://github.com/huggingface/peft/blob/df71b84341ae1ab3bc9b0d5f906d7a524850b63b/src/peft/tuners/lora.py#L539 they removed this from the original implementation

not sure whether we need to do the same.

lantiga:
Looks great, I propose we merge this one and keep the other in progress

finetune.py
```diff
@@ -73,19 +51,19 @@ def main() -> None:
     config = LLaMAConfig.from_name("7B")
     config.block_size = block_size
 
-    with fabric.device, with_lora(r=8, alpha=32, dropout=0.1, enabled=True):
+    with with_lora(r=lora_r, alpha=lora_alpha, dropout=lora_dropout, enabled=True):
         model = LLaMA(config)
+        print(model.transformer.h[0].attn.c_attn.weight.dtype)
+        print(type(model.transformer.h[0].attn.c_attn))
```
awaelchli, line 57:
```suggestion
```

PR #65, opened 2023-03-31
Fixes in finetune script

awaelchli:


lit_llama/lora.py
```diff
@@ -214,9 +214,8 @@ def __init__(self, config: llama.LLaMAConfig, rope_cache: torch.Tensor) -> None:
 
 
 @contextmanager
-def with_lora(r, alpha, dropout, enabled: bool = True):
-    """A context manager under which you can instantiate the model with LLoRA.
-    """
+def lora(r, alpha, dropout, enabled: bool = True):
+    """A context manager under which you can instantiate the model with LLoRA."""
```
lantiga, line 218:
```suggestion
    """A context manager under which you can instantiate the model with LoRA."""
```

lantiga (approved):
Looks good, I'd say let's err on the side of merging often

Issue #68, opened 2023-03-31
Vocabulary size when training the tokenizer


kno10:
For all I can tell, this will likely use the sentencepiece defaults (i.e., 8000 tokens vocabulary size) whereas Llama was supposedly trained using a vocabulary size of 32000?

https://github.com/Lightning-AI/lit-llama/blob/ba505cb18f66f7aa76bcef8520b9d45d8a145376/lit_llama/tokenizer.py#L46-L49

My suggestion is to
- [ ] add a parameter to set the vocabulary size
- [ ] set the default value to 32000 to match Llama?

lantiga:
Good point, in the tests we are loading the tokenizer from the checkpoint, but one should be able to set it from there as well.

Thanks, doing that rn

awaelchli:
Yes we could add that parameter. Note that we haven't actually used this train() function for llama, it is just there for the train.py skeleton script which for now runs on shakespeare, and tokenizer.train() is applied to that data. For the large amount of data that llama requires, one could anyway not feed all of it to this train() function. Just an FYI but otherwise the suggestion sounds good

kno10:
Fair enough - I'd assume that it makes more sense to pre-train the tokenizer independently anyway (on a manual sample of suitable size).

lantiga:
Done! Yes for Shakespeare I set it to a lower number explicitly. Thanks for opening the issue @kno10

PR #69, opened 2023-03-31
Add vocab_size to Tokenizer.train

lantiga:
Addresses #68

### DIFF
--- a/lit_llama/model.py
+++ b/lit_llama/model.py
@@ -13,12 +13,12 @@ from typing_extensions import Self
 
 
 def build_rope_cache(seq_len: int, n_elem: int, dtype: torch.dtype, base: int = 10000) -> torch.Tensor:
-    """This is the `build_rope_cache` implementation we initially intended to use, but it is numerically not
-    exactly equivalent to the one in the Meta model. We keep it here for posterity.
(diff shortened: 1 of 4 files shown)


lantiga:
BTW we should also add a test for tokenizer training, we are not covering that part right now.

lit_llama/model.py
```diff
@@ -13,12 +13,12 @@
 
 
 def build_rope_cache(seq_len: int, n_elem: int, dtype: torch.dtype, base: int = 10000) -> torch.Tensor:
-    """Enhanced Transformer with Rotary Position Embedding.
+    """This is the `build_rope_cache` implementation we initially intended to use, but it is numerically not
+    exactly equivalent to the one in the Meta model. We keep it here for posterity.
```
awaelchli, line 17: Going back to this implementation means we won't be able to load other weights directly and use the model. One could 1) add a flag that switches between the two implementation 2) have this one but then somehow adapt/correct them with tuning. Any other ideas?

lantiga: My bad, I messed up with my earlier edits. I wasn't trying to go back to the old rope implementation.

awaelchli (approved):

PR #81, opened 2023-04-02
Support DDP

### LINKED ISSUE #80: How to finetune with the multi-GPU
Hi, I'm wondering how to change the code for multi-GPU finetuning. Currently, I tried
`fabric = L.Fabric(devices=4, accelerator="gpu", strategy="ddp")`
But I encounter an error about initialization:
`RuntimeError: Default process group has not been initialized, please make sure to call init_process_group.`

awaelchli:
Fixes #80

Note, while we can enable DDP here, training/finetuning the model with DDP won't work on most systems as the model/optimizer/gradients simply can't fit in memory. This PR is just for correctness and because of the question that popped up in #80.

### DIFF
--- a/finetune.py
+++ b/finetune.py
@@ -37,6 +37,7 @@ warmup_steps = 100
 
 def main():
     fabric = L.Fabric(accelerator="cuda", devices=1)
+    fabric.launch()
     fabric.seed_everything(1337 + fabric.global_rank)
 
     if fabric.global_rank == 0:
--- a/lit_llama/model.py
+++ b/lit_llama/model.py
@@ -28,7 +28,8 @@ def build_rope_cache(seq_len: int, n_elem: int, dtype: torch.dtype, base: int = 
     # Calculate the product of position index and $\theta_i$
     idx_theta = torch.outer(seq_idx, theta)
 
-    return idx_theta
+    cache = torch.polar(torch.ones_like(idx_theta), idx_theta)  # complex64
+    return cache
 
 
 def apply_rope(x: torch.Tensor, rope_cache: torch.Tensor) -> torch.Tensor:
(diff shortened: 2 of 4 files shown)


lit_llama/model.py
```diff
@@ -28,17 +28,15 @@ def build_rope_cache(seq_len: int, n_elem: int, dtype: torch.dtype, base: int =
     # Calculate the product of position index and $\theta_i$
     idx_theta = torch.outer(seq_idx, theta)
 
-    # Cache them
-    cache = torch.polar(torch.ones_like(idx_theta), idx_theta)  # complex64
```
awaelchli, line 32: DDP doesn't support broadcasting complex64 buffers. 

```
    _sync_module_states(
  File "/home/adrian/.conda/envs/lit-llama/lib/python3.10/site-packages/torch/distributed/utils.py", line 142, in _sync_module_states
    _sync_params_and_buffers(
  File "/home/adrian/.conda/envs/lit-llama/lib/python3.10/site-packages/torch/distributed/utils.py", line 160, in _sync_params_and_buffers
    dist._broadcast_coalesced(
RuntimeError: Input tensor data type is not supported for NCCL process group: ComplexFloat
```

Solution: We can move the computation of polar() to the forward.

finetune.py
```diff
@@ -37,6 +37,7 @@
 
 def main():
     fabric = L.Fabric(accelerator="cuda", devices=1)
+    fabric.launch()
```
awaelchli, line 40: Add launch() here so user can just change devices above.

lit_llama/model.py
lantiga: Yes makes sense. We could even move computing the whole rope cache to forward.

lantiga: Which would also fix the failing test

lantiga:
I think the original request was referring to multi-gpu, be it ddp, fsdp or deepspeed.

I’m in favor of merging this after the rope fix, but I’d probably avoid showing ddp since it can’t possibly work, and rather make sure fsdp works instead.

WDYT?

lantiga (approved):

PR #85, opened 2023-04-03
Fix rope in LoRA-CausalSelfAttention

### LINKED ISSUE #83: Missing rope_cache for model with lora
Hi, according to the #81, the rope_cache argument has been removed from CausalSelfAttention in the model.py. I think the same process should also be done in the lora.py

awaelchli:
Fixes #83


lantiga (approved):
Looks great, I just starting to tackle this as well.

PR #89, opened 2023-04-03
Save checkpoint in train.py

### LINKED ISSUE #88: Saving FSDP Model
I was trying to change the fine-tuning to use FSDP training but currently there is any way to save the checkpoint.

Saving in [train.py](https://github.com/Lightning-AI/lit-llama/blob/main/train.py#L93) is commented and the saving in [finetune.pt](https://github.com/Lightning-AI/lit-llama/blob/ac83160ecaf9e0e55e0cb8fdbbd466d774c0a2f8/lit_llama/lora.py#L162) only supports Lora. 

I am trying to compare full fine-tuning with Lora. For full fine-tuning using FSDP my checkpoint is saved with the embedding layer and lm_head in a single `flat_param` which prevents me to load the checkpoint afterwards. How can I recover the original model architecture and load that checkpoint into a single GPU for inference ?

awaelchli:
Adds saving in train.py. The full consolidated checkpoint gets saved so it can be loaded into single-gpu inference. This won't work for the larger models. This is only temporary until Fabric supports saving and loading with FSDP. 

Note: If we don't like the boilerplate this adds to train.py, we can also hold off merging this and instead wait for Fabric to address this in general.

Fixes #88

### DIFF
--- a/train.py
+++ b/train.py
@@ -17,6 +17,8 @@ from torch.distributed.fsdp.wrap import transformer_auto_wrap_policy
 import numpy as np
 
 from lit_llama.model import Block, LLaMA, LLaMAConfig
+from lit_llama.utils import save_model_checkpoint
+
 
 out_dir = "out/training"
 eval_interval = 2000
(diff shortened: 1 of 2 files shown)


train.py
```diff
@@ -87,12 +87,21 @@ def train(
...
-        if iter_num > 0 and iter_num % eval_interval == 0 and fabric.global_rank == 0:
+        if iter_num > 0 and iter_num % eval_interval == 0:
             val_loss = validate(fabric, model, val_data)
             fabric.print(f"step {iter_num}: val loss {val_loss:.4f}")
-            # TODO: Save with Fabric
-            # print(f"saving checkpoint to {out_dir}")
-            # torch.save(checkpoint, os.path.join(out_dir, 'ckpt.pt'))
+
+            # TODO: Update this to `fabric.save()` once it supports FSDP checkpointing
+            from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
```
lantiga, line 95: let's maybe condition this on the strategy being FSDP
(I'm experimenting with training and I'm just training a smaller model without FDSP)

lantiga (changes requested):

train.py
awaelchli: That's a good idea, done that. Since boilerplate was getting a bit large for the train.py, I moved it to a utils.py file, is that fine?

lantiga (approved):
Looks good!
I would suggest adding a comment to the util that says that this will be upstreamed to Fabric soon.

Issue #87, opened 2023-04-03
How  to Convert the weights from Huggleface format to the Lit-LLaMA format


zhangsanfeng86:
Hi ,

Thank you for release code!

Could you share the script that How  to Convert the weights from Huggleface format to the Lit-LLaMA format?

lantiga:
Duplicate of #64 

We don't have the script right now but we'll be adding it asap

zhangsanfeng86:
Thanks for your reply!

timothylimyl:
@lantiga , is there a reverse? lit-llama to hugging face.

carmocca:
@timothylimyl Not at the moment. See https://github.com/Lightning-AI/lit-llama/issues/150

Issue #66, opened 2023-03-31
error when run  [python3 generate.py --quantize true ]: undefined symbol: cget_col_row_stats 


alexhmyang:
u20@u20:~/lit-llama$ python3 generate.py --quantize true --prompt "Hello, my name is"
/usr/lib/python3/dist-packages/requests/__init__.py:89: RequestsDependencyWarning: urllib3 (1.26.15) or chardet (3.0.4) doesn't match a supported version!
  warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported "
/home/u20/.local/lib/python3.8/site-packages/torchvision/io/image.py:13: UserWarning: Failed to load image Python extension: libtorch_cuda_cu.so: cannot open shared object file: No such file or directory
  warn(f"Failed to load image Python extension: {e}")
Loading model ...
Traceback (most recent call last):
  File "generate.py", line 147, in <module>
    CLI(main)
  File "/home/u20/.local/lib/python3.8/site-packages/jsonargparse/cli.py", line 82, in CLI
    return _run_component(component, cfg_init)
  File "/home/u20/.local/lib/python3.8/site-packages/jsonargparse/cli.py", line 138, in _run_component
    return component(**cfg)
  File "generate.py", line 108, in main
    model = LLaMA.from_name(model_size)
  File "/home/u20/lit-llama/lit_llama/model.py", line 223, in from_name
    return cls(LLaMAConfig.from_name(name))
  File "/home/u20/lit-llama/lit_llama/model.py", line 179, in __init__
    self.lm_head = nn.Linear(config.n_embd, config.vocab_size, bias=False)
  File "/home/u20/lit-llama/lit_llama/quantization.py", line 31, in __init__
    self._quantize_weight(self.weight.data)
  File "/home/u20/lit-llama/lit_llama/quantization.py", line 48, in _quantize_weight
    CB, CBt, SCB, SCBt, coo_tensorB = bnb.functional.double_quant(B)
  File "/home/u20/.local/lib/python3.8/site-packages/bitsandbytes/functional.py", line 1616, in double_quant
    row_stats, col_stats, nnz_row_ptr = get_colrow_absmax(
  File "/home/u20/.local/lib/python3.8/site-packages/bitsandbytes/functional.py", line 1505, in get_colrow_absmax
    lib.cget_col_row_stats(ptrA, ptrRowStats, ptrColStats, ptrNnzrows, ct.c_float(threshold), rows, cols)
[...]

Issue #64, opened 2023-03-30
How can I run it using Hugging Face models?


kadirnar:
Model: https://huggingface.co/decapoda-research/llama-7b-hf

awaelchli:
Hey @kadirnar 
You mean using the huggingface weights and load them into our model? This is theoretically possible, but one would have to write a function that remaps the state keys in the checkpoints to our format (and probably reshape/rearrange the tensors). In addition, since the hf model is also split across multiple files, you would have to merge them. So not a straightforward answer sorry :) We could think about providing such a conversion function for the short term if that's desired. I think that in long term though, our goal of the project is not to rely on the pretrained weights due to license restrictions.

lantiga:
I agree @awaelchli. To be practical, since I'm seeing people jumping on that checkpoint (before Meta takes it down 😅) we should provide a conversion script. Anyone wants to volunteer?

kadirnar:
@patrickvonplaten @osanseviero @julien-c

Can you look at this?

lantiga:
Looking into this @kadirnar

kadirnar:
> Looking into this @kadirnar

Thank you. Can you update the readme file? Can you explain step by step?

PR #90, opened 2023-04-03
Add conversion script for HF checkpoints

lantiga:
Addresses #64 #66 #87 

I'm getting close results, but not identical. I'm doing the reverse of `convert_llama_weights_to_hf.py` in the transformers repo, I suspect that the difference may be in RMSNorm or rotary embeddings.

### DIFF
--- a/scripts/convert_hf_checkpoint.py
+++ b/scripts/convert_hf_checkpoint.py
@@ -6,6 +6,8 @@ import torch
 from lit_llama.model import LLaMA, LLaMAConfig
 
 
+# Perform the reverse operation of: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/convert_llama_weights_to_hf.py
+
 def convert_hf_checkpoint(
     model_size: str = "7B", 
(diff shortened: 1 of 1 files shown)


scripts/convert_hf_checkpoint.py
```diff
@@ -0,0 +1,95 @@
+from pathlib import Path
+
+from transformers import LlamaForCausalLM
+import torch
+
+from lit_llama.model import LLaMA, LLaMAConfig
+
+
+# Perform the reverse operation of: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/convert_llama_weights_to_hf.py
```
awaelchli, line 9: Should this comment be attached to the function?

awaelchli (approved):
I think it would be nice if we could merge both scripts. Could be in a follow up if agreed.

lantiga:
> I think it would be nice if we could merge both scripts. Could be in a follow up if agreed.

Yep agreed.

scripts/convert_hf_checkpoint.py
lantiga: Done

PR #91, opened 2023-04-03
Support bfloat16 init

t-vi:
With this patch, I can init in bfloat16, this helps run 7B on my RTX consumer card.

### DIFF
--- a/lit_llama/utils.py
+++ b/lit_llama/utils.py
@@ -1,0 +1,36 @@
+import torch.utils._device
+
+
+class EmptyInitOnDevice(torch.overrides.TorchFunctionMode):
+    def __init__(self, device=None, dtype=None):
+        """
+        Create tensors with given device and dtype and don't run initialization
+           (but instead use "empty tensors", i.e. uninitialized memory).
+
+        Example::
+            with EmptyInitOnDevice("cuda", dtype=torch.bfloat16):
+               model = LLaMA.from_name('7B')
+            model.load_state_dict(torch.load('llama-lit/7B/state_dict.pth'))"""
+        self.device = device
+        self.dtype = dtype
+
+    def __torch_function__(self, func, types, args=(), kwargs=None):
+        kwargs = kwargs or {}
+        if getattr(func, "__module__", None) == "torch.nn.init":
+            if "tensor" in kwargs:
+                return kwargs["tensor"]
+            else:
+                return args[0]
+        if (
+            self.device is not None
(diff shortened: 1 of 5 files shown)


lantiga:
Great, I would have it in utils for now, and then we upstream to PL/Fabric

t-vi:
OK, so I'll add a bfloat16 command line option for generate tomorrow and update the readme...

t-vi:
If the CI doesn't complain, I think I'm reasonably happy with it.

lantiga (approved):
Looks great, let's hear from @awaelchli too

lit_llama/utils.py
```diff
@@ -0,0 +1,59 @@
...
+            with EmptyInitOnDevice("cuda", dtype=torch.bfloat16):
+               model = LLaMA.from_name('7B')
+            model.load_state_dict(torch.load('llama-lit/7B/state_dict.pth'))"""
+
+        self.quantization_mode = quantization_mode
+        if self.quantization_mode == 'llm.int8':
+            if device.type != "cuda":
+                raise ValueError("Quantization is only supported on the GPU.")
+            from .quantization import Linear8bitLt
+            self.Linear8bitLt = Linear8bitLt
```
awaelchli, line 25: shouldn't this be moved to `__enter__` directly?

lit_llama/utils.py
```diff
@@ -0,0 +1,59 @@
+import torch.utils._device
+
+
+class EmptyInitOnDevice(torch.overrides.TorchFunctionMode):
+    def __init__(self, device=None, dtype=None, quantization_mode=None):
+        """
+        Create tensors with given device and dtype and don't run initialization
+           (but instead use "empty tensors", i.e. uninitialized memory).
```
awaelchli, line 8: I don't see where the creation of empty tensors happens.

t-vi: PyTorch uses torch.empty to create the buffers for modules and then uses the overriden `toch.nn.init.*` to initialize:

https://github.com/pytorch/pytorch/blob/d47a4bf53fcde103ea6ea52eef66d999f707f61f/torch/nn/modules/linear.py#L96-L98

lit_llama/utils.py
t-vi: I have reservations importing in `__enter__` but maybe that is just my excellent taste in code style and cannot be generalized.

awaelchli (approved):

PR #92, opened 2023-04-04
Add GPTQ-style evaluation metrics (PPL on wikitext, ptb, c4)

t-vi:
We might or might not want to deduplicate instantiation between generate.py and evaluate.py, but I am not sure how.

For reference, I get (using bf16):
Perplexity on wikitext: 5.72
Perplexity on ptb: 8.38
Perplexity on c4: 7.32

They report perplexity of 5.68 for wikitext using fp16, so this is in line. (fp16, I cannot seem to reproduce, not sure if they mean AMP or so).

### DIFF
--- a/evaluate.py
+++ b/evaluate.py
@@ -1,0 +1,145 @@
+# This mimics GPTQ's evaluation metrics: https://github.com/IST-DASLab/gptq/
+# Thanks to E. Frantar et al GPTQ: Accurate Post-training Compression for GPT, arXiv:2210.17323
+import math
+import sys
+import time
+from pathlib import Path
+from typing import Optional
+
+import lightning as L
+import torch
+import tqdm
+
+from lit_llama import LLaMA, Tokenizer, as_8_bit_quantized
+from lit_llama.utils import EmptyInitOnDevice
+
+from datasets import load_dataset
+
+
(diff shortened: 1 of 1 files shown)


lantiga (approved):
Looks great, no need to de-duplicate.

evaluate.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+        encoded_text = encoded_text[
+            None, : 256 * model.config.block_size
+        ]  # add batch dimension, trim like gptq implementation
+        t0 = time.perf_counter()
+
+        nlls = 0
+        toks = 0
+        with torch.inference_mode():
+            bs = 2048  # this is for compat with gptq, and indeed we get much worse beyond this
```
awaelchli, line 114: It is not mentioned in the paper but I think this is simply because LLaMA was trained on a block size of 2048. And so that is why it could get worse for longer ones.

We could rename bs to block_size to avoid confusion with batch_size

awaelchli (approved):

evaluate.py
lantiga: Let's do that @t-vi

t-vi: Right, thank you!

PR #95, opened 2023-04-05
GPTQ int4 quantization

t-vi:
This adds int4 quantization, bringing the 7B model's GPU memory usage for generation to about ~5GB
The obvious next step is triton quantized matmul kernels to recover the performance.
Quantization takes <= 15 minutes when not disabling the TF32 matmuls (GPTQ upstream does this), I am not sure about the accuracy impact yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,14 +104,22 @@ This will run the 7B model and require ~26 GB of GPU memory (A100 GPU).
 
 ### Run Lit-LLaMA on consumer devices
 
-For GPUs with less memory, enable quantization (`--quantize true`) or use bfloat16 (`--dtype bfloat16`). Quantization will take longer to load but require ~8GB of memory. bfloat16 is closer to the "full deal" and runs on ~10GB of GPU memory.
+For GPUs with less memory, enable quantization (`--quantize llm.int8`) or use bfloat16 (`--dtype bfloat16`). Quantization will take longer to load but require ~8GB of memory. bfloat16 is closer to the "full deal" and runs on ~10GB of GPU memory.
 This can run on any consumer GPU.
 
 ```bash
-python generate.py --quantize true --prompt "Hello, my name is"
+python generate.py --quantize llm.int8 --prompt "Hello, my name is"
 ```
 
 See `python generate.py --help` for more options.
+
+You can also use GPTQ-style int4 quantization, but this needs conversions of the weights first:
+
+```bash
+python quantize.py --checkpoint_path state_dict.pth --tokenizer_path tokenizer.model --output_path llama-7b-gptq.4bit.pt --dtype bfloat16  --quantize gptq.int4
+```
+
+With the generated quantized checkpoint generation works as usual with `--quantize gptq.int4`, bringing GPU usage to about ~5GB. As only the weights of the Linear layers are quantized, it is useful to use `--dtype bfloat16` even with the quantization enabled.
 
 &nbsp;
 
@@ -163,6 +171,7 @@ Don't forget to [join our Discord](https://discord.gg/VptPCZkGNa)!
 - [@FacebookResearch](https://github.com/facebookresearch) for the original [LLaMA implementation](https://github.com/facebookresearch/llama)
 - [@TimDettmers](https://github.com/TimDettmers) for [bitsandbytes](https://github.com/TimDettmers/bitsandbytes)
 - [@Microsoft](https://github.com/microsoft) for [LoRA](https://github.com/microsoft/LoRA)
+- [@IST-DASLab](https://github.com/IST-DASLab) for [GPTQ](https://github.com/IST-DASLab/gptq)
 
 ## License
 
--- a/evaluate.py
+++ b/evaluate.py
@@ -10,7 +10,7 @@ import lightning as L
 import torch
 import tqdm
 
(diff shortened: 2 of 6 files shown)


quantize.py
```diff
@@ -0,0 +1,242 @@
...
+        groupsize=groupsize,
+        actorder=(groupsize == -1),
+    )
+    handle = model.lm_head.register_forward_hook(gptq.collect_input_stats)
+    for j in range(inps.size(0)):
+        model.lm_head(inps[j : j + 1])
+    handle.remove()
+    q_module, error = gptq.quantize()
+    model.lm_head = q_module
+    model._original_module.lm_head = q_module  # bleh.
```
t-vi, line 138: @awaelchli Ideas for this? If I don't I get the "original" linear module.

awaelchli: I think you may want to move

`model = fabric.setup_module(model)`
after
`llama_blockwise_quantization`

Otherwise, the model is wrapped already and that looks undesired here.

awaelchli: I just opened https://github.com/Lightning-AI/lightning/issues/17285 with an idea to solve this so the user doesn't have to think about it.

lantiga: @t-vi good to merge, but do you want to try moving `model = fabric.setup_module(model)` after `llama_blockwise_quantization` as suggested before merging?

t-vi: What would be the difference to commenting out? The only step left after the quantization is to save the state dict.

lantiga: Probably no difference at this point. We are passing `fabric` to save_model_checkpoint` but I don't think we are leveraging any fabric-specific functionality here.

At this point I would just remove the use of fabric from `quantize.py` altogether, as it's not needed.

t-vi: OK

quantize.py
```diff
@@ -0,0 +1,244 @@
...
+
+import lightning as L
+import torch
+import tqdm
+
+from lit_llama import LLaMA, Tokenizer
+from lit_llama.utils import EmptyInitOnDevice, save_model_checkpoint
+from lit_llama.quantization import GPTQQuantizer
+
+from datasets import load_dataset
```
awaelchli, line 19: we would need to add this to requirements.txt I think

quantize.py
```diff
@@ -0,0 +1,244 @@
...
+from pathlib import Path
+from typing import Optional
+import gc
+
+import lightning as L
+import torch
+import tqdm
+
+from lit_llama import LLaMA, Tokenizer
+from lit_llama.utils import EmptyInitOnDevice, save_model_checkpoint
```
awaelchli, line 16:
```suggestion
from lit_llama.utils import EmptyInitOnDevice
```
unused import :)

quantize.py
```diff
@@ -0,0 +1,244 @@
...
+    dtype: Optional[str] = None,
+    quantize: Optional[str] = None,
+) -> None:
+    """Generates text samples based on a pre-trained LLaMA model and tokenizer.
+
+    Args:
+        datasets: The datasets to use as a comma separated string
+        # compile: Whether to compile the model.
+        accelerator: The hardware to run on. Possible choices are:
+            ``"cpu"``, ``"cuda"``, ``"mps"``, ``"gpu"``, ``"tpu"``, ``"auto"``.
```
awaelchli, line 161:
```suggestion
```

quantize.py
```diff
@@ -0,0 +1,244 @@
...
+    model.lm_head = q_module
+    model.lm_head.to("cpu")
+
+
+def main(
+    datasets: str = "wikitext,ptb,c4",
+    *,
+    # compilation fails as it does not support torch.complex64 for RoPE
+    # compile: bool = False,
+    accelerator: str = "auto",
```
awaelchli, line 146:
```suggestion
```

quantize.py
```diff
@@ -0,0 +1,244 @@
...
+    q_module, error = gptq.quantize()
+    model.lm_head = q_module
+    model.lm_head.to("cpu")
+
+
+def main(
+    datasets: str = "wikitext,ptb,c4",
+    *,
+    # compilation fails as it does not support torch.complex64 for RoPE
+    # compile: bool = False,
```
awaelchli, line 145:
```suggestion
```
I think torch.compile is not relevant for this script

quantize.py
```diff
@@ -0,0 +1,244 @@
...
+        t0 = time.time()
+        model = LLaMA.from_name(model_size)
+        checkpoint = torch.load(checkpoint_path)
+        model.load_state_dict(checkpoint)
+        print(f"Time to load model: {time.time() - t0:.02f} seconds.", file=sys.stderr)
+
+    model.eval()
+
+    # if compile:
+    #     model = torch.compile(model)
```
awaelchli, line 209:
```suggestion
```

awaelchli (approved):
Only minor nits 🎉

quantize.py
lantiga: ok let's do it

lantiga:
Merging, let's add datasets to requirements.txt in a separate micro PR :-)

lantiga:
Amazing contribution @t-vi !

t-vi:
Thanks @lantiga @awaelchli !

PR #106, opened 2023-04-06
Parity test for LLaMA-Adapter implementation

awaelchli:
Adds a small test that compares our implementation with the reference (model forward). 

TODO: 
- [x] Put original implementation in gist and download only for test


lit_llama/adapter.py
```diff
@@ -94,7 +94,7 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
             ak = ak.view(1, aT, self.n_head, head_size).repeat(B, 1, 1, 1).transpose(1, 2)
             av = av.view(1, aT, self.n_head, head_size).repeat(B, 1, 1, 1).transpose(1, 2)
 
-            ascores = torch.matmul(q, ak.transpose(2, 3)) / math.sqrt(self.n_embd)
+            ascores = torch.matmul(q, ak.transpose(2, 3)) / math.sqrt(head_size)
```
awaelchli, line 97: This was a small bug. Doesn't break the adaption but the scaling here lead to an inconsistency. I didn't notice this previously in my comparisons because in order for this to be noticable, the gate needs to be > 0 otherwise adaption is effectively disabled.

tests/test_model.py
```diff
@@ -107,3 +107,73 @@ def test_bfloat16_llama_init(lit_llama, orig_llama) -> None:
...
+    llama_model.apply(llama_model._init_weights)
+    orig_llama_model = orig_llama.Transformer(orig_llama_config)
+
+    copy_weights(llama_model, orig_llama_model)
+    copy_adapter_weights(llama_model, orig_llama_model)
+
+    # make the gate non-zero, otherwise the adapter is disabled and the model
+    # identical to regular LLaMA
+    enable_gate(llama_model)
+    enable_gate(orig_llama_model)
```
awaelchli, line 175: It took my slow brain a while to realize this fact

lit_llama/adapter.py
lantiga: great catch, I actually looked at this last night and it escaped it too

tests/test_model.py
lantiga: lol

lantiga (approved):
Great job @awaelchli

PR #108, opened 2023-04-06
batch hf conversion to save ram

factoidforrest:
This does appear to work but I haven't figured out how to test it yet and I seem to be getting OOM on the verify step still, even on the 7b model and 60gb of ram. It says "Killed" and I'm not sure how to debug it. 

If verification is disabled, this will use about 30gb of ram on 7B, which is ideal as many folks will have 32gb. 

Anyway, this should get you much closer and again it does appear to work.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ numpy  # train.py dataset memmap
 jsonargparse[signatures]  # generate.py, convert_checkpoint.py CLI
 bitsandbytes  # quantization.py
 datasets  # evaluate.py
+git+https://github.com/huggingface/transformers
--- a/scripts/convert_hf_checkpoint.py
+++ b/scripts/convert_hf_checkpoint.py
@@ -1,14 +1,18 @@
 from pathlib import Path
-
+import sys
 # support running without installing as a package
 wd = Path(__file__).parent.parent.resolve()
 sys.path.append(str(wd))
 
-from transformers import LlamaForCausalLM
 import torch
-
 from lit_llama.model import LLaMA, LLaMAConfig
+import os
+import json
+from transformers import LlamaForCausalLM
 
+def load_weights_from_bin(weight_path: Path, weight_file: str) -> torch.Tensor:
+    weights = torch.load(weight_path / weight_file, map_location="cpu")
+    return weights
 
 def convert_hf_checkpoint(
     model_size: str = "7B",
@@ -16,6 +20,8 @@ def convert_hf_checkpoint(
     lit_checkpoint: Path = Path("checkpoints/lit-llama.ckpt"),
     verify: bool = False,
 ) -> None:
+
+
     """
     Perform the reverse operation of: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/convert_llama_weights_to_hf.py
     """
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -6,3 +6,4 @@ numpy  # train.py dataset memmap
 jsonargparse[signatures]  # generate.py, convert_checkpoint.py CLI
 bitsandbytes  # quantization.py
 datasets  # evaluate.py
+git+https://github.com/huggingface/transformers
```
lantiga, line 9: I'd avoid adding transformers to the dependencies since it's only used during verification, and instead produce a useful error message ("verification requires transformers to be installed, please pip install ..." during verification if you don't have it.

scripts/convert_hf_checkpoint.py
```diff
@@ -1,37 +1,54 @@
...
 wd = Path(__file__).parent.parent.resolve()
 sys.path.append(str(wd))
 
-from transformers import LlamaForCausalLM
 import torch
-
 from lit_llama.model import LLaMA, LLaMAConfig
+import os
+import json
+from transformers import LlamaForCausalLM
```
lantiga, line 11: Let's move the import to the verification scope (line 121)

scripts/convert_hf_checkpoint.py
```diff
@@ -41,53 +58,81 @@ def permute(w):
...
+                        sd["transformer.wte.weight"].copy_(hf_weights["model.embed_tokens.weight"])
+
+                    if "model.norm.weight" in hf_weights:
+                        sd["transformer.ln_f.scale"].copy_(hf_weights["model.norm.weight"])
+
+                    if "lm_head.weight" in hf_weights:
+                        sd["lm_head.weight"].copy_(hf_weights["lm_head.weight"])
+
+                    # Break here to avoid reloading the same weights multiple times
+                    # this assumes the layers won't be split between bin files, which hopefully is not possible
```
lantiga, line 116: the original llama weights are indeed split between the files (because it was tensor-parallel), but it is not the case with HF models AFAIK

lantiga (commented):
Great job!
A couple of changes requests (related to imports), then we can merge.
In the meantime I'll test the verification of a beefy machine

lantiga:
I have follow-up changes that will allow working in bfloat16 directly saving 50% memory, so let's land this one and I'll layer the other on top of it

lantiga:
BTW I just tested verification and unfortunately it fails. I'm now testing the same on main with the 13B model.

lantiga:
I tested this on my end: the new conversion script is producing garbage in output, so there's still something missing there

Previous
```
Here is what people think about pineapple pizza.
The pineapple pizza is a pizza with the traditional Italian topping of tomato, mozzarella and basil, seasoned with oregano and garlic and topped with pineapple chunks.
```

Current
```
Here is what people think about pineapple pizza М Vernpertiesхи Americanspertiesissantlaw becameа fueron instance/).ціалькто worry octдра校хи directive Mitchell Quick Theripeline TextView moabaseético Bosluss season insert SDInvalidartifact coff%%%%%%%% FOR votreборculesдраadeshзыва Quickcules staat moิ
```

scripts/convert_hf_checkpoint.py
factoidforrest: got it, so we should be ok with this assumption then?

factoidforrest:
> I tested this on my end: the new conversion script is producing garbage in output, so there's still something missing there
> 
> Previous
> 
> ```
> Here is what people think about pineapple pizza.
> The pineapple pizza is a pizza with the traditional Italian topping of tomato, mozzarella and basil, seasoned with oregano and garlic and topped with pineapple chunks.
> ```
> 
> Current
> 
> ```
> Here is what people think about pineapple pizza М Vernpertiesхи Americanspertiesissantlaw becameа fueron instance/).ціалькто worry octдра校хи directive Mitchell Quick Theripeline TextView moabaseético Bosluss season insert SDInvalidartifact coff%%%%%%%% FOR votreборculesдраadeshзыва Quickcules staat moิ
> ```

oh shoot! I wonder what is going wrong. Maybe a missing layer. How do I take this checkpoint output and run inference on it? This script produces a .chkp file which is tensorflow stuff but your inference code takes .pth. Is the file extension maybe just misnamed?

lantiga:
Hey the extension is just very fluid (I’ve seen pt pth ckpt…)
just pass the ckpt file to generate.py and it’ll work

lantiga:
Hey @factoidforrest I'm merging to a local branch (`low-ram-hf`) so we can fix this directly

PR #112, opened 2023-04-07
Use flash attention in adapters by moving gating factor out of matmul

### LINKED ISSUE #107: Use FlashAttention with LLaMA-Adapter
Looking at the LLaMA-Adapter implementation, at https://github.com/Lightning-AI/lit-llama/blob/main/lit_llama/adapter.py#L91

```python
            # inefficient attention because we need to insert the gate for the adaption in the middle
            aT = prefix.size(1)
            _, ak, av = self.c_attn(prefix).split(self.n_embd, dim=2)
            ak = ak.view(1, aT, self.n_head, head_size).repeat(B, 1, 1, 1).transpose(1, 2)
            av = av.view(1, aT, self.n_head, head_size).repeat(B, 1, 1, 1).transpose(1, 2)

            ascores = torch.matmul(q, ak.transpose(2, 3)) / math.sqrt(self.n_embd)
            ascores = self.gating_factor * F.softmax(ascores.float(), dim=-1).type_as(q)
            y = y + torch.matmul(ascores, av)
```

it looks to me we could replace the above with

```python
            a_mask = torch.ones(aT, aT, dtype=torch.bool)
            ay = F.scaled_dot_product_attention(q, ak, av, attn_mask=a_mask, dropout_p=0.0, is_causal=False)
            y = y + self.gating_factor * ay
```

since
```
(gating * softmax(ascores)) @ av
```
is equivalent to
```
gating * (softmax(ascores) @ av)
```

lantiga:
Fixes #107, needs #106 for verification

### DIFF
--- a/lit_llama/adapter.py
+++ b/lit_llama/adapter.py
@@ -88,15 +88,14 @@ class CausalSelfAttention(nn.Module):
         if self.block_idx >= self.adapter_start_layer:
             prefix = self.adapter_wte.weight.reshape(1, self.adapter_prompt_length, self.n_embd)
 
-            # inefficient attention because we need to insert the gate for the adaption in the middle
             aT = prefix.size(1)
             _, ak, av = self.c_attn(prefix).split(self.n_embd, dim=2)
             ak = ak.view(1, aT, self.n_head, head_size).repeat(B, 1, 1, 1).transpose(1, 2)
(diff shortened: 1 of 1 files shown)


lantiga:
Whoops, fixing the test now

lantiga:
Done

awaelchli (approved):
🎉

PR #114, opened 2023-04-10
Update prepare_alpaca.py to enable data_file_name specification

mosheber:


lantiga (approved):
It doesn't hurt to have the additional flexibility, since there are multiple sources of Alpaca data IMO

PR #100, opened 2023-04-05
Finetuning with weights in bfloat16

awaelchli:
The memory consumption is now ~20GB compared to before at ~38GB. 
The training iteration speed-up ~1.5x-2x.

PR #117, opened 2023-04-10
Fix weight initialization in LoRA and Adapter finetuning

### LINKED ISSUE #116: Convergence of LLaMA-adapter
Dear Sir,

I find that the implementation of LLaMa-adapter can not converge. The loss function keeps around 0.8-1.0.

I wonder if you can solve this?

And thanks for your attention.

awaelchli:
Fixes #116

In #100 I accidentally applied the EmptyInitOnDevice context manager over the adapter/LoRA weights.

### DIFF
--- a/finetune_adapter.py
+++ b/finetune_adapter.py
@@ -55,8 +55,10 @@ def main():
 
     checkpoint = torch.load("checkpoints/lit-llama/7B/state_dict.pth")
 
-    with EmptyInitOnDevice(device=fabric.device, dtype=torch.bfloat16):
-        model = LLaMA(config)
+    with fabric.device:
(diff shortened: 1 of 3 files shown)


finetune_adapter.py
```diff
@@ -55,8 +55,10 @@ def main():
 
     checkpoint = torch.load("checkpoints/lit-llama/7B/state_dict.pth")
 
-    with EmptyInitOnDevice(device=fabric.device, dtype=torch.bfloat16):
-        model = LLaMA(config)
+    with fabric.device:
+        torch.set_default_tensor_type(torch.HalfTensor)
```
lantiga, line 59: Not for this PR, but we could take this pattern out to a context so users don't have to remember to do the dance

PR #118, opened 2023-04-11
Raise exception if checkpoints are not found

lantiga:


scripts/convert_checkpoint.py
```diff
@@ -92,6 +92,9 @@ def meta_weights_for_nano_model(
     checkpoint_files.sort()
     n_checkpoints = len(checkpoint_files)
 
+    if n_checkpoints == 0:
+        raise RuntimeError(f"No checkpoints were found at ckpt_dir ${ckpt_dir}. `consolidated.0*.pth` files expected at that location.")
```
awaelchli, line 96: is the dollar sign intentional or a typo?

awaelchli (approved):

scripts/convert_checkpoint.py
lantiga: typo, thanks!

PR #122, opened 2023-04-12
Batch hf conversion to save ram

lantiga:
Continuation of #108 by @factoidforrest

It also supports converting to a given `dtype`, e.g. for `bfloat16` (if you omit it will convert to `float32`):

```
python scripts/convert_hf_checkpoint.py --hf_checkpoint_path checkpoints/llama-hf/llama-13b-hf --lit_checkpoint lit-llama.pth --model_size 13B --dtype bfloat16
```

It's much faster than before and it uses way less memory.

### DIFF
--- a/evaluate.py
+++ b/evaluate.py
@@ -78,6 +78,8 @@ def main(
         if not isinstance(dt, torch.dtype):
             raise ValueError(f"{dtype} is not a valid dtype.")
         dtype = dt
+    else:
+        dtype = torch.float32
 
     with EmptyInitOnDevice(
         device=fabric.device, dtype=dtype, quantization_mode=quantize
--- a/evaluate.py
+++ b/evaluate.py
@@ -48,7 +48,7 @@ def main(
     checkpoint_path: Optional[Path] = None,
     tokenizer_path: Optional[Path] = None,
     model_size: str = "7B",
-    dtype: Optional[str] = None,
+    dtype: str = "float32",
     quantize: Optional[str] = None,
 ) -> None:
     """Generates text samples based on a pre-trained LLaMA model and tokenizer.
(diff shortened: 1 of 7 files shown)


evaluate.py
```diff
@@ -78,6 +78,8 @@ def main(
         if not isinstance(dt, torch.dtype):
             raise ValueError(f"{dtype} is not a valid dtype.")
         dtype = dt
+    else:
```
carmocca, line 81: In L77 you can replace the default return `None` with `torch.float32` and remove this else block

This suggestion applies to the other files too

scripts/convert_hf_checkpoint.py
```diff
@@ -41,58 +58,85 @@ def permute(w):
...
+        "model.norm.weight": "transformer.ln_f.scale",
+        "lm_head.weight": "lm_head.weight"
+    }
+
+    for bin_file in bin_files:
+        print("Processing", bin_file)
+
+        hf_weights = torch.load(os.path.join(hf_checkpoint_path, bin_file), map_location="cpu")
+
+        with torch.no_grad():
```
carmocca, line 83: Convert this context manager to a function decorator so everything is covered. This will also remove it from inside the for loop

scripts/convert_hf_checkpoint.py
```diff
@@ -41,58 +58,85 @@ def permute(w):
...
-                sd_hf[f"model.layers.{i}.mlp.down_proj.weight"]
-            )
-
-            sd[f"transformer.h.{i}.rms_1.scale"].copy_(
-                sd_hf[f"model.layers.{i}.input_layernorm.weight"]
-            )
-            sd[f"transformer.h.{i}.rms_2.scale"].copy_(
-                sd_hf[f"model.layers.{i}.post_attention_layernorm.weight"]
-            )
+    bin_files = set(el for el in bin_index["weight_map"].values())
```
carmocca, line 61: nit: move this beside where you load the `bin_index`, above `def permute`

carmocca (approved):
Great!

evaluate.py
```diff
@@ -78,6 +78,8 @@ def main(
...
+        dtype = torch.float32
```
carmocca, line 82: If you change `dtype` to `: str = "float32` in the function definition, we could get rid of this `if/else` and also make it clearer what's the default

This comment applies to the other files too

lantiga: Done!

scripts/convert_hf_checkpoint.py
lantiga: Done

scripts/convert_hf_checkpoint.py
lantiga: Good point, done

PR #124, opened 2023-04-12
Enable checkpoint with FSDP on single device

### LINKED ISSUE #113: Cannot save checkpoint while train.py on single GPU
I got following error at `utils.py` while saving checkpoint while pre-training on single GPU. Any hint how should I fix it? Thanks.

```
state_dict = model._forward_module.state_dict()                            

NotImplementedError: offload_to_cpu=True and NO_SHARD is not supported yet
```

awaelchli:
Fixes  #113

In #113, the user wants to run the training with FSDP but with a single device. FSDP does not support offloading the checkpoint state to CPU if world-size is 1.

### DIFF
--- a/lit_llama/utils.py
+++ b/lit_llama/utils.py
@@ -15,7 +15,7 @@ def save_model_checkpoint(fabric, model, file_path):
     """
 
     if isinstance(fabric.strategy, FSDPStrategy):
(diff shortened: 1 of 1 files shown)


lantiga (approved):
Great!

PR #128, opened 2023-04-12
Improved finetune_adapter.py

awaelchli:
A few changes to the finetune_adapter.py that made it successful for me. 

- Fixes a bug where the eos token was inserted at the end of the input sequence in evaluation. 
- Adds deepspeed zero-2 with gradient accumulation to train in <= 1 hour. Once Fabric supports checkpointing with FSDP, we can switch over so users don't have to install an additional package.
- In a follow up, I will add a function that converts the deepspeed checkpoint so it can be easily loaded by the inference script. Right now you need to run `python path/to/checkpoint/zero_to_fp32.py` by hand.


finetune_adapter.py
```diff
@@ -15,33 +19,45 @@
...
-batch_size = 64
-micro_batch_size = 4
+batch_size = 64 / devices
+micro_batch_size = 8
 gradient_accumulation_steps = batch_size // micro_batch_size
 epoch_size = 50000  # train dataset size
-num_epochs = 100
-max_iters = epoch_size * num_epochs // micro_batch_size  # 5 epochs
+num_epochs = 5
+max_iters = num_epochs * epoch_size // devices  # 5 epochs
```
carmocca, line 42: Already above

```suggestion
max_iters = num_epochs * epoch_size // devices
```

finetune_adapter.py
```diff
@@ -53,6 +69,8 @@ def main():
     config = LLaMAConfig()
     config.block_size = block_size
 
+    # If you get an error here, check that you have downloaded the weights by following the
+    # instructions in the README.
```
carmocca, line 73: Maybe assert `path.isfile()` and point to the README if it fails

generate.py
```diff
@@ -156,4 +157,8 @@ def main(
     from jsonargparse import CLI
 
     torch.set_float32_matmul_precision("high")
+    warnings.filterwarnings(
```
carmocca, line 160: Can you add a comment that mentions which component is triggering this warning?

generate_adapter.py
```diff
@@ -0,0 +1,137 @@
...
+
+    output = output[:eos_pos]
+    return output
+
+
+if __name__ == "__main__":
+    from jsonargparse import CLI
+
+    torch.set_float32_matmul_precision("high")
+    warnings.filterwarnings(
```
carmocca, line 133: Same suggestion

carmocca (approved):

generate.py
awaelchli: Added. It comes from the empty tensor (through our use of EmptyInitOnDevice)

finetune_adapter.py
```diff
@@ -4,6 +4,10 @@
 LLaMA-Adapter: Efficient Fine-tuning of Language Models with Zero-init Attention
 https://arxiv.org/abs/2303.16199
 
+This script uses DeepSpeed Zero-2 to train efficiently on 8 A100 GPUs within 1 hour as done in the original paper.
+If you have fewer GPUs, you can remove the `DeepSpeedStrategy` from Fabric and set `devices = 1`.
```
lantiga, line 8: The script already sets strategy to `None` if devices are <2, so probably this comment should be changed to reflect that

lantiga (approved):
Looks great, kudos @awaelchli

finetune_adapter.py
awaelchli: Okay I added the error message. But I suggest we only make an exception here. In general let's keep the validation light in favor of minimalism.

PR #132, opened 2023-04-12
DeepSpeed checkpoint consolidation

awaelchli:
Consolidates the deepspeed checkpoint to a single file for easier loading at inference (at inference we don't require deepspeed to be set up). 

User no longer needs to convert the checkpoint manually. The generate_adapter.py script will work without extra steps needed after running finetune_adapter.py


lantiga (approved):
Approved! We could enable saving in half precision as well, so it can be loaded more easily later, but we can leave this for a follow-up PR.

awaelchli:
The checkpoint already is in half-precision. The deepspeed function you see here doesn't care about the dtype, despite its name including fp32.

PR #133, opened 2023-04-13
Improved LoRA finetuning

### LINKED ISSUE #77: Question: about left padding
if the model is padded to the left and its a casual language model does that mean that padding tokens will receive attention from the rest of the sequence?

Should there me a mask to prevent tokens to attend to padding?

awaelchli:
Small fixes carried over from finetune_adapter.py
- right padding
- remove redundant gradient clipping
- avoid tensor cloning warning from torch

Fixes #77


lantiga (approved):
Looks good!

carmocca (approved):

PR #136, opened 2023-04-13
Update URL of cleaned alpaca data

dspinellis:


scripts/prepare_alpaca.py
```diff
@@ -14,7 +14,7 @@
 from tqdm import tqdm
 
 
-DATA_FILE = "https://raw.githubusercontent.com/tloen/alpaca-lora/main/alpaca_data_cleaned.json"
+DATA_FILE = "https://raw.githubusercontent.com/tloen/alpaca-lora/main/alpaca_data_cleaned_archive.json"
 DATA_FILE_NAME = "alpaca_data_cleaned.json"
```
carmocca, line 18: This needs to change too

dspinellis: Indeed, it can be changed, but it doesn't have to be changed, and I don't think it should be changed.  The `DATA_FILE_NAME` is used locally to name the data file. There is no need for it to reflect the remote file name, especially if the remote name was changed to reflect the other project's operations (the fact that it has been archived).

carmocca: Sounds good. Thank you for your contribution!

carmocca (approved):

awaelchli (approved):
Thanks!

PR #143, opened 2023-04-16
[WIP] add dolly dataset script

aniketmaurya:


awaelchli:
Do you think it would be possible to merge the dolly finetune script with the regular one? I don't see any drastic differences between the two. In https://github.com/Lightning-AI/lit-llama/pull/144 I made the weights path and data path configurable, maybe that's already enough?

lantiga:
Thank you @aniketmaurya! I agree with @awaelchli 

Rather than introducing a new script, you could actually add a howto that describes the process end to end.
In the future we will start rely on howto's to describe a complete journey with code that can be cut and pasted.

Of course the current finetuning howto's potentially already cover 90% of it, but it would be great to have a howto dedicated to Dolly. Same goes for future toolformer etc tasks.

aniketmaurya:
thanks @awaelchli @lantiga for the feedback, I will merge the finetuning script. I can add how-to doc in a follow up PR.

lantiga:
The finetuning script should be exactly the same right?

finetune/finetune_lora_dolly.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def generate_response(model, instruction):
+    # TODO: make it configurable
+    tokenizer = Tokenizer("../../weights/tokenizer.model")
+    sample = {"instruction": instruction, "context": "", "category": "open_qa"}
+    prompt = generate_prompt(sample)
+    encoded = tokenizer.encode(prompt, bos=True, eos=False)
+    encoded = encoded[None, :]  # add batch dimension
+    encoded = encoded.to(model.device)
```
carmocca, line 132: Beware of #162

scripts/prepare_dolly.py
```diff
@@ -0,0 +1,133 @@
...
+    file_path = destination_path / DATA_FILE_NAME
+    download(file_path)
+
+    # TODO: If we don't have the Meta weights, where do we get the tokenizer from?
+    tokenizer = Tokenizer(tokenizer_path)
+
+    with open(file_path, "r") as file:
+        data = file.readlines()
+        data = [json.loads(line) for line in data]
+    for item in data:
```
aniketmaurya, line 46: the original dolly dataset contains - `instruction`, `context`, `category` and `response` keys.

I have mapped context-> input and response -> output based on the Alpaca dataset. `category` contains the info of the type of instruction and response. Not sure how valuable this information could be for fine-tuning, but we can append this in the prompt type of task.

After preparing the data using this script, users can directly use the `finetune_lora.py` script by providing the path to this dataset.

lantiga: Not sure about `category`, maybe we can look at the original Dolly codebase whether they include it during finetuning?

scripts/prepare_dolly.py
```diff
@@ -0,0 +1,133 @@
...
+    
+    The output is a training and validation dataset saved as `train.pt` and `val.pt`,
+    which stores the preprocessed and tokenized prompts and labels.
+    """
+    
+    destination_path.mkdir(parents=True, exist_ok=True)
+    file_path = destination_path / DATA_FILE_NAME
+    download(file_path)
+
+    # TODO: If we don't have the Meta weights, where do we get the tokenizer from?
```
lantiga, line 40: This is a good question.
We should probably distribute the tokenizer model ourselves in the future, to make this less painful. It's 220k gzipped, and I don't think anyone would complain, no real IP in there.

scripts/prepare_dolly.py
aniketmaurya: @lantiga I quickly checked their prompt for training [here](https://github.com/databrickslabs/dolly/blob/master/training/consts.py) and couldn't find an usage of category.


On reading their [blog](https://www.databricks.com/blog/2023/04/12/dolly-first-open-commercially-viable-instruction-tuned-llm), I find that the categories were created for 7 tasks and rewarding the labelers:

```
We set up a contest, where the top 20 labelers would get a big award. We also outlined 7 very specific tasks:
```

scripts/prepare_dolly.py
lantiga: Not for this PR of course

lantiga (approved):
Two quick checks and we're good to go!

scripts/prepare_dolly.py
awaelchli: It is not used for training: https://github.com/search?q=repo%3Adatabrickslabs%2Fdolly%20category&type=code

scripts/prepare_dolly.py
```diff
@@ -0,0 +1,133 @@
...
+DATA_FILE = "https://raw.githubusercontent.com/databrickslabs/dolly/master/data/databricks-dolly-15k.jsonl"
+DATA_FILE_NAME = "dolly_data_cleaned.json"
+IGNORE_INDEX = -1
+
+
+def prepare(
+    destination_path: Path = Path("data/dolly"), 
+    tokenizer_path: Path = Path("checkpoints/lit-llama/tokenizer.model"),
+    test_split_size: int = 2000,
+    max_seq_length: int = 256,
```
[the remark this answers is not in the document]
aniketmaurya: I looked at [their code](https://github.com/databrickslabs/dolly/blob/0eadcb7b0648d496d67243a7d572b413560be661/training/trainer.py#L223) and couldn't find the config for `max_seq_length`. Their fallback if not specified during training is 1024.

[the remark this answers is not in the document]
lantiga: Ok let's go with 1024 then

lantiga, line 26:
```suggestion
    max_seq_length: int = 1024,
```

lantiga:
Merging as soon as tests pass

lantiga:
Merged, great job @aniketmaurya !

PR #145, opened 2023-04-17
Add generation script for LoRA

lantiga:
Doesn't support `quantize` yet.

### DIFF
--- a/generate_lora.py
+++ b/generate_lora.py
@@ -66,7 +66,7 @@ def main(
     assert tokenizer_path.is_file()
 
     if quantize is not None:
-        raise NotImplemented("Quantization in LoRA is not supported yet")
+        raise NotImplementedError("Quantization in LoRA is not supported yet")
 
     fabric = L.Fabric(accelerator=accelerator, devices=1)
 
--- a/generate_lora.py
+++ b/generate_lora.py
@@ -65,6 +65,9 @@ def main(
     assert pretrained_path.is_file()
     assert tokenizer_path.is_file()
 
+    if quantize is not None:
+        raise NotImplemented("Quantization in LoRA is not supported yet")
+
     fabric = L.Fabric(accelerator=accelerator, devices=1)
 
     dt = getattr(torch, dtype, None)
--- a/generate_lora.py
+++ b/generate_lora.py
@@ -86,7 +86,7 @@ def main(
         pretrained_checkpoint = torch.load(pretrained_path)
         model.load_state_dict(pretrained_checkpoint, strict=False)
 
-        # 2. Load the fine-tuned adapter weights
+        # 2. Load the fine-tuned LoRA weights
         lora_checkpoint = torch.load(lora_path)
         model.load_state_dict(lora_checkpoint, strict=False)
 
--- a/generate_lora.py
+++ b/generate_lora.py
@@ -55,7 +55,7 @@ def main(
             ``"cpu"``, ``"cuda"``, ``"mps"``, ``"gpu"``, ``"tpu"``, ``"auto"``.
     """
     if not lora_path:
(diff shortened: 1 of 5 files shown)


generate_lora.py
```diff
@@ -0,0 +1,149 @@
...
+    temperature: float = 0.8,
+    accelerator: str = "auto",
+) -> None:
+    """Generates a response based on a given instruction and an optional input.
+    This script will only work with checkpoints from the instruction-tuned LoRA model.
+    See `finetune_lora.py`.
+
+    Args:
+        prompt: The prompt/instruction (Alpaca style).
+        lota_path: Path to the checkpoint with trained LoRA weights, which are the output of
```
awaelchli, line 41:
```suggestion
        lora_path: Path to the checkpoint with trained LoRA weights, which are the output of
```

generate_lora.py
```diff
@@ -0,0 +1,149 @@
...
+import warnings
+from pathlib import Path
+from typing import Optional
+
+import lightning as L
+import torch
+
+from generate import generate
+from lit_llama import Tokenizer
+from lit_llama.adapter import LLaMA, LLaMAConfig
```
awaelchli, line 12: The import needs to be fixed

generate_lora.py
```diff
@@ -0,0 +1,149 @@
...
+        dtype: The dtype to use during generation.
+        max_new_tokens: The number of generation steps to take.
+        top_k: The number of top most probable tokens to consider in the sampling process.
+        temperature: A value controlling the randomness of the sampling process. Higher values result in more random
+            samples.
+        accelerator: The hardware to run on. Possible choices are:
+            ``"cpu"``, ``"cuda"``, ``"mps"``, ``"gpu"``, ``"tpu"``, ``"auto"``.
+    """
+    if not lora_path:
+        lora_path = Path("out/adapter/alpaca/lit-llama-lora-finetuned.pth")
```
awaelchli, line 58:
```suggestion
        lora_path = Path("out/lora/alpaca/lit-llama-lora-finetuned.pth")
```
And the finetune_lora is a bit different from finetune-alpaca as we don't write an end-of-trianing checkpoint

generate_lora.py
```diff
@@ -0,0 +1,149 @@
...
+        pretrained_path = Path(f"./checkpoints/lit-llama/7B/lit-llama.pth")
+    if not tokenizer_path:
+        tokenizer_path = Path("./checkpoints/lit-llama/tokenizer.model")
+    
+    assert lora_path.is_file()
+    assert pretrained_path.is_file()
+    assert tokenizer_path.is_file()
+
+    if quantize is not None:
+        raise NotImplementedError("Quantization in LoRA is not supported yet")
```
awaelchli, line 69: Should we remove the argument then? In adapter it is also not supported atm.

generate_lora.py
```diff
@@ -0,0 +1,149 @@
...
+    ), lora(r=lora_r, alpha=lora_alpha, dropout=lora_dropout, enabled=True):
+        print("Loading model ...", file=sys.stderr)
+        t0 = time.time()
+        model = LLaMA(LLaMAConfig())  # TODO: Support different model sizes
+
+        # 1. Load the pretrained weights
+        pretrained_checkpoint = torch.load(pretrained_path)
+        model.load_state_dict(pretrained_checkpoint, strict=False)
+
+        # 2. Load the fine-tuned adapter weights
```
awaelchli, line 89:
```suggestion
        # 2. Load the fine-tuned LoRA weights
```

awaelchli (approved):

generate_lora.py
lantiga: I did add saving in finetune_alpaca.py

generate_lora.py
lantiga: I kept it since it's probably the next thing we need to look into

PR #151, opened 2023-04-17
Save and load only adapter weights

awaelchli:
- Save only adapter weights for smaller checkpoints and easier sharing: This requires a bit of inefficient hacking for the case we are running finetuning with multi-gpu deepspeed. We should look for ways to make this simpler in the future.
- Load adapter weights separately for inference: This requires a small fix for our custom bnb.Linear layer. 

Addresses https://github.com/Lightning-AI/lit-llama/pull/149#discussion_r1169028988

Note: LoRA finetuning already has this covered.

### DIFF
--- a/finetune_adapter.py
+++ b/finetune_adapter.py
@@ -13,15 +13,15 @@ Note: If you run into a CUDA error "Expected is_sm80 to be true, but got false",
 """
 import os
 import time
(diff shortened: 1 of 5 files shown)


lantiga (approved):
Looks great!

Issue #49, opened 2023-03-28
Doesn't work with batched input


patrickvonplaten:
It seems like the generate method doesn't work with batched input no - it only accepts an input of type `str`

awaelchli:
@patrickvonplaten 

The generate method inside of generate.py accepts a batched input. The docstring says:

```py
idx: Tensor of shape (B, T) with indices of the prompt sequence.
```
This is the tokenized version of the string.

You can actually see how it gets passed to this method here: https://github.com/Lightning-AI/lit-llama/blob/9810b07ce74a811c3682cee5be6b0d3da15877aa/generate.py#LL122C6-L122C6 

where we just add the batch dimension. The generate.py is meant to be run as a simple script with a simple prompt as input from the command line like shown in the README. But if you want, you can also import the function and use it for a batched input. 

Alternatively, one could also move the tokenization into the generate function and make it accept a list of strings.

lantiga:
Ok to close this @patrickvonplaten ?

carmocca:
> Alternatively, one could also move the tokenization into the generate function and make it accept a list of strings.

We could add on more function for this that basically contains this piece of code: https://github.com/Lightning-AI/lit-llama/blob/f808df1971cbac345b3d8de263a251374a53d5e9/generate.py#L120-L136 with support for multiple prompts

But if we feel like the change would add unnecessary misdirection, then we can close this issue.

patrickvonplaten:
Ok to close, just note that batched generation probably also need things like left-padding, e.g. if samples are of different lengths, then the padding should be added on the let

`[["hey"], ["hey this is"]]` => `[[0, 0, 34], [34, 2, 3]]`

Might be worth adding some helping util for this, but ok to close

lantiga:
Closing as we removed batch generation support to make the generation code simpler.
Happy to add a separate batched generation script later on if the need emerges.

PR #153, opened 2023-04-18
Stop generating once <eos> token is triggered

### LINKED ISSUE #130: Terminate early at EOS when performing inference
Our inference script (generate.py) and instruct.py (if #128 lands) currently generate text up to the max length. However, the model can generate an EOS token any time. We should stop the sequence and truncate it. In #128 this truncation was done after generating the full sequence, but we could already early stop during [the for loop](https://github.com/Lightning-AI/lit-llama/blob/094fa95523dc3f157aa357c8d1beaee6377f4459/generate.py#L42).

timothylimyl:
Hi there, 

I think we should just stop generating any new tokens once the <eos> token is found. The user can specify to trigger this feature by providing the <eos> token id to the `generate()` function via the argument `eos_id`.  We only need this feature for instruction tuned models, so I just added the `eos_id` argument into lora and adapter generate function respectively.

Have a good day!

Fixes #130

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -19,6 +19,7 @@ def generate(
     max_seq_length: int,
     temperature: float = 1.0,
     top_k: Optional[int] = None,
+    eos_id: Optional[int] = None,
 ) -> torch.Tensor:
     """Takes a conditioning sequence (prompt) as input and continues to generate as many tokens as requested.
 
@@ -57,9 +58,13 @@ def generate(
 
         probs = torch.nn.functional.softmax(logits, dim=-1)
         idx_next = torch.multinomial(probs, num_samples=1)
-
+        
+        # if <eos> token is triggered, return the output (stop generation)
+        if idx_next==eos_id:
+            return idx[:,:t]
+        
         # concatenate the new column
-        idx[:, t:] = idx_next
+        idx[:, t] = idx_next
 
     return idx
 
--- a/generate.py
+++ b/generate.py
@@ -32,6 +32,7 @@ def generate(
         max_seq_length: The maximum sequence length allowed.
         temperature: Scales the predicted logits by 1 / temperature
         top_k: If specified, only sample among the tokens with the k highest probabilities
+        eos_id: If specified, stop generating any more token once the <eos> token is triggered
     """
     # create an empty tensor of the expected final shape and fill in the current tokens
     B, T = idx.shape
--- a/generate.py
+++ b/generate.py
@@ -65,7 +65,7 @@ def generate(
 
         # if <eos> token is triggered, return the output (stop generation)
         if idx_next == eos_id:
-            return idx[:t]
+            return idx[:t + 1]  # include the EOS token
 
     return idx
 
--- a/generate.py
+++ b/generate.py
@@ -27,7 +27,7 @@ def generate(
 
     Args:
         model: The model to use.
-        idx: Tensor of shape (B, T) with indices of the prompt sequence.
+        idx: Tensor of shape (T) with indices of the prompt sequence.
         max_new_tokens: The number of new tokens to generate.
         max_seq_length: The maximum sequence length allowed.
         temperature: Scales the predicted logits by 1 / temperature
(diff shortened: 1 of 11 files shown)


lantiga (approved):
Looks great, thanks for the contribution @timothylimyl !

lantiga:
Wait, talked too soon, there are failures in the tests :-)

generate.py
```diff
@@ -57,9 +59,13 @@ def generate(
 
         probs = torch.nn.functional.softmax(logits, dim=-1)
         idx_next = torch.multinomial(probs, num_samples=1)
-
+        
+        # if <eos> token is triggered, return the output (stop generation)
+        if idx_next==eos_id:
```
carmocca, line 64: This is batched so this wont work. See my comment at https://github.com/Lightning-AI/lit-llama/issues/130#issuecomment-1505564783

lantiga:
Weird, I can't repro the failure locally

carmocca:
@lantiga The command to run is `pytest tests/test_generate.py::test_generate[2] -v`. Maybe you missed checking out this branch

generate.py
lantiga: Ah, of course you’re right.
this speeds up generation, so I’d vouch for following this approach

lantiga:
Possible, I had multiple repos / terminals open when I gave it a shot

timothylimyl:
Sorry about that, I was not aware that we are aiming to generate multiple batches of prompts. In all the generation scripts, the prompt is of type `str` rather than `list[str]`. Please let me know what fixes can I do,  I'm excited to contribute.

awaelchli:
We can also drop the batch support, we don't use it anywhere in our demo scripts. It is just there as an artifact because we adapted the code from nanoGPT. IMO we should do it because it would simplify the code and a batched version of the function can always be added if needed.

carmocca:
If we remove it, then https://github.com/Lightning-AI/lit-llama/issues/49 should be closed

timothylimyl:
I reckon that the fixes be done here first in `tokenizer.py` if we want to support batch generation:

```
    def encode(
        self,
        string: str,
        bos: bool = True,
        eos: bool = False,
        max_length: int = -1,
        pad: bool = False,
        device: Optional[torch.device] = None
    ) -> torch.Tensor:
        tokens = self.processor.encode(string)
        if bos:
            tokens = [self.bos_id] + tokens
        if eos:
            tokens = tokens + [self.eos_id]
        if max_length > 0:
            tokens = tokens[:max_length]
        if pad and len(tokens) < max_length:
            tokens += [self.pad_id] * (max_length - len(tokens))

        return torch.tensor(tokens, dtype=torch.int, device=device)
```

I think we can take in a list of strings as default, if the length is one then add a batch dimension, if the length is above one then we loop to concatenate the `tokens` on the first dim (batch dim).  

Something along these lines (quick pseudocode):
```
    output_tokens_tensor = []
    for string in list_of_prompts:
        tokens = self.processor.encode(string)
        if bos:
            tokens = [self.bos_id] + tokens
        if eos:
            tokens = tokens + [self.eos_id]
        if max_length > 0:
            tokens = tokens[:max_length]
        if pad and len(tokens) < max_length:
            tokens += [self.pad_id] * (max_length - len(tokens))
        
        output_tokens_tensor.append(token)
    
    output_tokens_tensor = torch.stack(output_tokens_tensor)
    return output_tokens_tensor.to(device)

 
```

In `generate.py`, we can then make it:
```
    encoded_prompt = tokenizer.encode(prompt, bos=True, eos=False, device=fabric.device)
    encoded_prompt = encoded_prompt[None, :]  # add batch dimension
```
to 
```
    encoded_prompt = tokenizer.encode(prompt, bos=True, eos=False, device=fabric.device)
```

Then I will fix the generate function to deal with batches.  Let me know if you think this is worth looking into.

timothylimyl:
> We can also drop the batch support, we don't use it anywhere in our demo scripts. It is just there as an artifact because we adapted the code from nanoGPT. IMO we should do it because it would simplify the code and a batched version of the function can always be added if needed.

I can help out with this in the future for the next PR?

timothylimyl:
Okay, so are we finalising going with the current approach of single batch generation first? Anything to fix in this PR?

timothylimyl:
> @timothylimyl I suggest that you remove batch support in a separate PR. When that's merged, the changes here will work

Hi @carmocca , a little confused here on what changes to make (this PR is meant to work for the current version of a single user prompt).

carmocca:
@timothylimyl There is a test for batched generation (the one failing!), and the function implements a `B` dimension which should be removed (also docstrings etc)

My proposal is that you do this in a separate PR

carmocca:
@timothylimyl I opened #162 to unblock this

carmocca (approved):

timothylimyl:
> @timothylimyl I opened #162 to unblock this

awesome, thank you!

lantiga (approved):
Stamped! Thank you @timothylimyl!

awaelchli (commented):
Needs #166 
Otherwise LGTM

PR #157, opened 2023-04-18
Add Red Pajama conversion script for sample and full dataset

### LINKED ISSUE #155: Add training script that loads RedPajama data

lantiga:
Fixes #155

### DIFF
--- a/lit_llama/indexed_dataset.py
+++ b/lit_llama/indexed_dataset.py
@@ -1,0 +1,588 @@
+# https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/data/indexed_dataset.py
+
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of the FairSeq source tree.
+
+# copied from fairseq/fairseq/data/indexed_dataset.py
+# Removed IndexedRawTextDataset since it relied on Fairseq dictionary
+# other slight modifications to remove fairseq dependencies
+# Added document index to index file and made it accessible.
+#    An empty sentence no longer separates documents.
+
+from functools import lru_cache
+import os
+import shutil
+import struct
+from itertools import accumulate
+
+import numpy as np
+import torch
+
+
+def __best_fitting_dtype(vocab_size=None):
+    if vocab_size is not None and vocab_size < 65500:
+        return np.uint16
+    else:
+        return np.int32
+
+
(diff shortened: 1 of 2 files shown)


lantiga:
We'll need to start testing these scripts by mocking some stuff out.

lantiga:
I'll add a howto in a subsequent PR

lit_llama/indexed_dataset.py
```diff
@@ -0,0 +1,588 @@
...
+
+class MMapIndexedDataset(torch.utils.data.Dataset):
+    class Index(object):
+        _HDR_MAGIC = b'MMIDIDX\x00\x00'
+
+        @classmethod
+        def writer(cls, path, dtype):
+            class _Writer(object):
+                def __enter__(self):
+                    self._file = open(path, 'wb')
```
awaelchli, line 348: There is a class defined inside a class, and inside the writer method of that inner class, it defines another class.

lantiga:
Ready to review @awaelchli @carmocca

scripts/prepare_redpajama.py
```diff
@@ -0,0 +1,183 @@
...
+
+        filepath = source_path / name
+
+        if not filepath.is_file():
+            raise RuntimeError(
+                f"Input file not found at {filepath}. \n"
+                "Make sure you download the data, e.g. wget -i https://data.together.xyz/redpajama-data-1T/v1.0.0/urls.txt or through \n"
+                "https://huggingface.co/datasets/togethercomputer/RedPajama-Data-1T \n"
+                "https://huggingface.co/datasets/togethercomputer/RedPajama-Data-1T-Sample \n"
+            )
```
awaelchli, line 67: For me, this error didn't trigger. It just generated empty files in the output folder data/red_pajama_sample/.

awaelchli (approved):

scripts/prepare_redpajama.py
awaelchli: If glob didn't return files, this for loop will never run and then the `if not filepath.is_file():` check will never trigger

lit_llama/indexed_dataset.py
lantiga: Ahah, that's right. I took this from megatron (see file header) and basically didn't touch it.
I agree we could make it more readable.

scripts/prepare_redpajama.py
lantiga: Updated

PR #160, opened 2023-04-18
Automatically infer model from state_dict

### LINKED ISSUE #158: Emit informative error message when model size and state_dict don't match

dnhkng:
Fixes #158 

This adds a dictionary to the model (almost the inverse of llama_configs), and looks at the first layer of the loaded state_dict to infer the model type and load it.

### DIFF
--- a/evaluate.py
+++ b/evaluate.py
@@ -10,7 +10,7 @@ import lightning as L
 import torch
 import tqdm
 
-from lit_llama import LLaMA, Tokenizer
+from lit_llama import LLaMA, Tokenizer, llama_model_lookup
 from lit_llama.utils import EmptyInitOnDevice
 
 from datasets import load_dataset
@@ -85,7 +85,7 @@ def main(
         t0 = time.time()
         checkpoint = torch.load(checkpoint_path)
         model = LLaMA.from_name(
-            LLaMA.llama_model_lookup[checkpoint["lm_head.weight"].shape[1]]
+            llama_model_lookup[checkpoint["lm_head.weight"].shape[1]]
         )
         model.load_state_dict(checkpoint)
         print(f"Time to load model: {time.time() - t0:.02f} seconds.", file=sys.stderr)
--- a/evaluate.py
+++ b/evaluate.py
@@ -84,9 +84,8 @@ def main(
         print("Loading model ...", file=sys.stderr)
         t0 = time.time()
         checkpoint = torch.load(checkpoint_path)
-        model = LLaMA.from_name(
-            llama_model_lookup[checkpoint["lm_head.weight"].shape[1]]
-        )
+        name = llama_model_lookup(checkpoint)
+        model = LLaMA.from_name(name)
         model.load_state_dict(checkpoint)
         print(f"Time to load model: {time.time() - t0:.02f} seconds.", file=sys.stderr)
 
--- a/evaluate.py
+++ b/evaluate.py
@@ -10,8 +10,8 @@ import lightning as L
 import torch
 import tqdm
 
-from lit_llama import LLaMA, Tokenizer, llama_model_lookup
-from lit_llama.utils import EmptyInitOnDevice
+from lit_llama import LLaMA, Tokenizer
+from lit_llama.utils import EmptyInitOnDevice, llama_model_lookup
 
 from datasets import load_dataset
 
--- a/evaluate.py
+++ b/evaluate.py
@@ -47,7 +47,6 @@ def main(
     accelerator: str = "auto",
     checkpoint_path: Optional[Path] = None,
     tokenizer_path: Optional[Path] = None,
-    model_size: str = "7B",
     dtype: str = "float32",
     quantize: Optional[str] = None,
 ) -> None:
@@ -65,7 +64,7 @@ def main(
             ``"gptq.int4"``: GPTQ 4-bit mode.
     """
     if not checkpoint_path:
-        checkpoint_path = Path(f"./checkpoints/lit-llama/{model_size}/lit-llama.pth")
+        checkpoint_path = Path(f"./checkpoints/lit-llama/7B/lit-llama.pth")
     if not tokenizer_path:
         tokenizer_path = Path("./checkpoints/lit-llama/tokenizer.model")
     assert checkpoint_path.is_file()
--- a/evaluate.py
+++ b/evaluate.py
@@ -83,8 +83,10 @@ def main(
     ):
         print("Loading model ...", file=sys.stderr)
         t0 = time.time()
-        model = LLaMA.from_name(model_size)
         checkpoint = torch.load(checkpoint_path)
+        model = LLaMA.from_name(
+            LLaMA.llama_model_lookup[checkpoint["lm_head.weight"].shape[1]]
+        )
         model.load_state_dict(checkpoint)
         print(f"Time to load model: {time.time() - t0:.02f} seconds.", file=sys.stderr)
 
--- a/generate.py
+++ b/generate.py
@@ -7,7 +7,7 @@ from typing import Optional
 import lightning as L
 import torch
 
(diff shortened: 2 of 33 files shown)


dnhkng:
Not sure that a raw dict is the correct way to store the data, but it's the most straightforward.

However, it's a bit messy with the imports. How about we move the dict to a class variable of LLaMA, so we can access like LLaMA.from_name?

lit_llama/model.py
```diff
@@ -32,6 +32,12 @@ def from_name(cls, name: str) -> Self:
     "65B": LLaMAConfig(n_layer=80, n_head=64, n_embd=8192),
 }
 
+llama_model_lookup = {
```
lantiga, line 35: Add comment describing what this is and what the size refers to

lit_llama/__init__.py
```diff
@@ -1,2 +1,2 @@
-from lit_llama.model import LLaMAConfig, LLaMA, RMSNorm, build_rope_cache, apply_rope
+from lit_llama.model import LLaMAConfig, LLaMA, RMSNorm, llama_model_lookup, build_rope_cache, apply_rope
```
lantiga, line 1: I would not expose `llama_model_lookup` here. I would just import it from `lit_llama.utils` where needed.

evaluate.py
```diff
@@ -83,8 +83,10 @@ def main(
     ):
         print("Loading model ...", file=sys.stderr)
         t0 = time.time()
-        model = LLaMA.from_name(model_size)
         checkpoint = torch.load(checkpoint_path)
+        model = LLaMA.from_name(
+            llama_model_lookup[checkpoint["lm_head.weight"].shape[1]]
```
lantiga, line 88: I would turn `llama_model_lookup` into a function in `utils` that takes the whole checkpoint and returns the name. This way we can encapsulate the complete logic there, and just call

```
name = llama_model_lookup(checkpoint)
model = LLaMA.from_name(name)
```

lit_llama/model.py
lantiga: also, this will become a function

lantiga (changes requested):
A few changes but it looks good

lantiga:
Also, can you please add the same logic to `generate_lora.py` and `generate_adapter.py` as well? Thank you!

lit_llama/__init__.py
dnhkng: Seems reasonable, done!

evaluate.py
dnhkng: done

evaluate.py
```diff
@@ -10,7 +10,7 @@
 import torch
 import tqdm
 
-from lit_llama import LLaMA, Tokenizer
+from lit_llama import LLaMA, Tokenizer, llama_model_lookup
 from lit_llama.utils import EmptyInitOnDevice
```
awaelchli, line 14:
```suggestion
from lit_llama import LLaMA, Tokenizer
from lit_llama.utils import EmptyInitOnDevice, llama_model_lookup
```

lit_llama/utils.py
```diff
@@ -11,6 +11,27 @@
...
+    8192: "65B",  # 65B n_embd=8192
+}
+
+
+def llama_model_lookup_from_size(size: int) -> str:
+    """Returns the LLaMA model name from the embedding size."""
+    return llama_model_sizes[size]
+
+
+def llama_model_lookup(checkpoint: dict) -> str:
```
awaelchli, line 27: the two function could be combined.

generate_adapter.py
```diff
@@ -65,12 +65,14 @@ def main(
     ):
         print("Loading model ...", file=sys.stderr)
         t0 = time.time()
-        model = LLaMA(LLaMAConfig())  # TODO: Support different model sizes
 
         # 1. Load the pretrained weights
         pretrained_checkpoint = torch.load(pretrained_path, map_location=torch.device("cpu"))
+        name = llama_model_lookup(pretrained_checkpoint)
+        model = LLaMA.from_name(name)
```
awaelchli, line 72: This isn't supported for the adapter version of LLaMA.
Let's do `LLaMAConfig.from_name()` instead here, and pass in the config.

evaluate.py
dnhkng: ooops, missed a file ☺️

generate_adapter.py
lantiga: I'm not sure I understand why it's not supported though. Aren't we loading the unchanged frozen weights?

awaelchli: I'm just referring to the "from_name" function which either needs to be added or be avoided. Note that the LLaMA from model.py and LLaMA from adapter.py is different and thus code can easily get out of sync.

awaelchli: I can send a PR to rectify this inconvenience.

awaelchli: I'm addressing it here: #170
Hopefully the last thing to unblock this PR :)

dnhkng:
Can we get the CI/CD to run finetune_lora and finetune_adapter, and run generate for each of the artifacts to catch bugs early?

generate.py
```diff
@@ -113,11 +114,13 @@ def main(
...
-    checkpoint = lazy_load(checkpoint_path)
-    model.load_state_dict(checkpoint)
-    print(f"Time to load model: {time.time() - t0:.02f} seconds.", file=sys.stderr)
+        print("Loading model ...", file=sys.stderr)
+        t0 = time.time()
+        checkpoint = lazy_load(checkpoint_path)
+        name = llama_model_lookup(checkpoint)
+        model = LLaMA.from_name(name)
+        model.load_state_dict(checkpoint)
+        print(f"Time to load model: {time.time() - t0:.02f} seconds.", file=sys.stderr)
```
awaelchli, line 123: moving this inside here is reverting https://github.com/Lightning-AI/lit-llama/commit/88ba00804a63684a819fb4960fef778d3900a8ee

so we need to adjust this

awaelchli: same applies to the other generate scripts as well.

generate.py
```diff
@@ -113,11 +114,13 @@ def main(
     with EmptyInitOnDevice(
         device=fabric.device, dtype=dtype, quantization_mode=quantize
     ):
-        model = LLaMA.from_name(model_size)
```
carmocca, line 116: We need to decide if we want to remove the `model_size` argument entirely now

generate.py
dnhkng: this block of code (or very similar versions) is repeated too many times:

```
    print("Loading model ...", file=sys.stderr)
    t0 = time.time()
    with EmptyInitOnDevice(
    ):
        model = LLaMA(LLaMAConfig())  # TODO: Support different model sizes

    # 1. Load the pretrained weights
    pretrained_checkpoint = lazy_load(pretrained_path)
    model.load_state_dict(pretrained_checkpoint, strict=False)
```
Can we add it to the utils module, and pass a path, and return the model?

dnhkng: OK, I have refactored model loading into utils.load_model
It can handle a regular model by default, or an adapter model when adapter=True.

awaelchli: I think we should definitely keep the loading statements in the individual scripts. They are intentionally duplicated for flexibility between the files and there are minor differences that matter. For example the LLaMA model for adapter is not the same as the regular LLaMA. To your question what is "lora", is a context manager that converts the layers in the LLaMA model to LoRA layers corresponding to the finetuned model from finetune_lora.py.

In any case, I think you achieved what was desired for #158 and I would keep further refactoring efforts out of this PR.

dnhkng: OK, I have reverted the load function and tidied it up.

lantiga (approved):
Looks great! Good to see friction go.
Last tweak is updating the README or other locations where we use `model_size` for generation.

PR #162, opened 2023-04-19
Remove batched generation support

### LINKED ISSUE #49: Doesn't work with batched input

carmocca:
Unblocks #153 
Closes #49

### DIFF
--- a/finetune_adapter.py
+++ b/finetune_adapter.py
@@ -156,9 +156,7 @@ def generate_response(model, instruction, input=""):
     tokenizer = Tokenizer("checkpoints/lit-llama/tokenizer.model")
     sample = {"instruction": instruction, "input": input}
     prompt = generate_prompt(sample)
-    encoded = tokenizer.encode(prompt, bos=True, eos=False)
-    encoded = encoded[None, :]  # add batch dimension
-    encoded = encoded.to(model.device)
+    encoded = tokenizer.encode(prompt, bos=True, eos=False, device=model.device)
 
     output = generate(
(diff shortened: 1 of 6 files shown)


finetune_adapter.py
```diff
@@ -167,7 +165,7 @@ def generate_response(model, instruction, input=""):
         max_new_tokens=100,
         temperature=0.8,
     )
-    output = tokenizer.decode(output[0].cpu())
```
carmocca, line 170: AFAIK `.cpu()` is not required. `Tensor.tolist()` works on a CUDA tensor

lantiga (approved):
Looks good, thanks @carmocca!
We can always add a separate script supporting batch generation in the future.

generate.py
```diff
@@ -26,40 +26,40 @@ def generate(
...
     for t in range(T, T_new):
         # ignore the not-filled-yet tokens
-        idx_cond = idx[:, :t]
+        idx_cond = idx[:t]
         # if the sequence context is growing too long we must crop it at max_seq_length
-        idx_cond = idx_cond if T <= max_seq_length else idx_cond[:, -max_seq_length:]
+        idx_cond = idx_cond if T <= max_seq_length else idx_cond[-max_seq_length:]
 
         # forward
         logits = model(idx_cond)
```
awaelchli, line 50: The model still expects a batch, so the input tensor needs to be expanded here.

PR #163, opened 2023-04-19
quantization: instantiate model on CPU

### LINKED ISSUE #148: OOM on VRAM during quantization
I just found a (potential) issue when quantizing the 13B+ models. Thing start out correctly and the first layer start quantizing correctly, but after reaching the level zero mlp level, I get on OOM error:
`0 mlp.c_fc2 collecting stats quantizing time 22s quantization error 351.9`
I get an out-of-memory error.

```
OutOfMemoryError: CUDA out of memory. Tried to allocate 730.00 MiB (GPU 0; 31.75 GiB total capacity; 29.65 GiB already allocated; 82.69 MiB free; 30.68 GiB reserved in total by PyTorch) 
If reserved memory is >> allocated memory try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF
```

Earlier I didn't get far enough, due to problems now resolved from issue #141 

Although I can in theory quantize 13B, in practice I can't even with 40Gb VRAM. I can't even start quantizing the 30B model, as I get OOM instantly.

I am using the command:
`python quantize.py --model_size 30B --checkpoint_path checkpoints/lit-llama/30B/lit-llama.pth --tokenizer_path checkpoints/lit-llama/tokenizer.model --output_path llama-30b-gptq.4bit.pt --dtype bfloat16  --quantize gptq.int4`

This is after using the new push, so the models are already in BF16 (the 30B model is now only 65 GB). I will try testing on our DGX on A100s, but I think it would be good to understand the high memory use.

t-vi:
Initialize the mode on CPU to save GPU RAM.

Fixes #148

### DIFF
--- a/quantize.py
+++ b/quantize.py
@@ -187,8 +187,9 @@ def main(
     else:
         raise RuntimeError(f"unknown/unsupported quantization mode {quantize}")
 
+    # we avoid loading the entire model on the GPU and do this block by block
     with EmptyInitOnDevice(
-        device=device,
+        device="cpu",
         dtype=dtype,
     ):
         print("Loading model ...", file=sys.stderr)


lantiga (approved):
Stamped!

PR #164, opened 2023-04-19
Add lazyload from torchhacks

t-vi:
...and use it in generate.

Drawback (7B model and default settings):
```
  Before: Time to load model: 4.01 seconds.
  After:  Time to load model: 4.14 seconds.
```

Advantage (from time -v):
```
  Before: Maximum resident set size (kbytes): 14103952
  After:  Maximum resident set size (kbytes): 1179236
```

The other thing here is copying vs. using torchhacks as a dependency, I went with the former for now, either way is OK with me.
Obviously, this depends on PyTorch internals more than other bits.

Are there places where we do not want to use lazy_load for the model loading? What about the datasets in `finetune*.py`?

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -8,7 +8,7 @@ import lightning as L
 import torch
 
 from lit_llama import LLaMA, Tokenizer
-from lit_llama.utils import EmptyInitOnDevice
+from lit_llama.utils import EmptyInitOnDevice, lazy_load
 
 
 @torch.no_grad()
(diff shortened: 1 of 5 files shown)


lantiga:
Thank you @t-vi ! Maybe we could provide a `low_mem` flag so you don't pay loading time if you don't have to?
Also it might be interesting to profile and see where the extra time is spent in the end.

t-vi:
OK, so the slowdown is gone if we use PyTorch's decode-zip-into-storage-directly.

lantiga:
Amazing @t-vi

lantiga:
@t-vi saw the quantization fix. Just to align, can we now add the trick to the other `generate_` scripts?

t-vi:
Yes, they do!

t-vi:
So do we want more things to use this or go with the generation for now?
Other than that, I'm reasonably happy with it now.

lantiga (approved):
Stamped!

lantiga:
> So do we want more things to use this or go with the generation for now?
> Other than that, I'm reasonably happy with it now.

It would be great to get this into the finetune scripts as well as a separate PR

lantiga:
Merging, really cool work @t-vi

t-vi:
ahahahaa. Sorry about that!

PR #168, opened 2023-04-19
Iterate only a single time per layer when converting the weights

LaurentMazare:
When converting the llama checkpoint to the weight file, the current code iterates over each layer multiple times as it loops over all the tensors that are in the state-dict. This simple change ensures that instead we loop over each `layer_idx` exactly once hence reducing the extraction time significantly.

### DIFF
--- a/scripts/convert_checkpoint.py
+++ b/scripts/convert_checkpoint.py
@@ -24,7 +24,7 @@ def convert_state_dict(state_dict: Dict[str, torch.Tensor], dtype: torch.dtype =
     converted["lm_head.weight"] = state_dict["output.weight"].to(dtype)
     converted["transformer.ln_f.scale"] = state_dict["norm.weight"].to(dtype)
 
-    for layer_idx in set([k.split(".")[1] for k in state_dict if k.startswith("layers")]):
+    for layer_idx in sorted(set([k.split(".")[1] for k in state_dict if k.startswith("layers")])):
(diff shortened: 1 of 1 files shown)


scripts/convert_checkpoint.py
```diff
@@ -24,9 +24,7 @@ def convert_state_dict(state_dict: Dict[str, torch.Tensor], dtype: torch.dtype =
     converted["lm_head.weight"] = state_dict["output.weight"].to(dtype)
     converted["transformer.ln_f.scale"] = state_dict["norm.weight"].to(dtype)
 
-    for key in [k for k in state_dict if k.startswith("layers")]:
-        layer_idx = key.split(".")[1]
-
+    for layer_idx in set([k.split(".")[1] for k in state_dict if k.startswith("layers")]):
```
awaelchli, line 27: iteration over the set is non-deterministic. in general, in case someone needs to step through the code with a debugger, it might be better to iterate over a sorted list here. but fine with the current approach too.

awaelchli (approved):

scripts/convert_checkpoint.py
LaurentMazare: Thanks for the quick feedback, I've added a `sorted` call to ensure that  the order is easier to reproduced.

awaelchli:
Thanks for this!

PR #170, opened 2023-04-19
Make `adapter.LLaMA.from_name` available

awaelchli:
We initially didn't expect that we would finetune the bigger models with adapters as well, so we only included the default 7B config. 

The PR #160 wants to call `LLaMA.from_name` but it is not available for the adapter implementation. This PR extends the existing config to enable this.

### DIFF
--- a/lit_llama/model.py
+++ b/lit_llama/model.py
@@ -22,14 +22,14 @@ class LLaMAConfig:
 
     @classmethod
     def from_name(cls, name: str) -> Self:
-        return llama_configs[name]
+        return cls(**llama_configs[name])
 
 
 llama_configs = {
-    "7B": LLaMAConfig(n_layer=32, n_head=32, n_embd=4096),
-    "13B": LLaMAConfig(n_layer=40, n_head=40, n_embd=5120),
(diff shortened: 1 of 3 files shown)


tests/test_adapter.py
```diff
@@ -0,0 +1,22 @@
+import torch
+from dataclasses import asdict
+import pytest
+
+
+@pytest.mark.parametrize("model_size", ["7B", "13B", "30B", "65B"])
+def test_config_identical(model_size, lit_llama):
```
awaelchli, line 7: Test isn't doing much but it's honest work and will fail should the definitions get out of sync somehow

lit_llama/model.py
```diff
@@ -22,14 +22,14 @@ class LLaMAConfig:
 
     @classmethod
     def from_name(cls, name: str) -> Self:
-        return llama_configs[name]
+        return cls(**llama_configs[name])
 
 
 llama_configs = {
```
carmocca, line 28: This could become a class attribute

carmocca (approved):

lantiga (approved):
Looks good. It's also cleaner overall, I like it

PR #169, opened 2023-04-19
Update howto guide to download weights

carmocca:
Fixes #147

PR #172, opened 2023-04-19
Add howto on customizing paths

carmocca:
Follow-up to https://github.com/Lightning-AI/lit-llama/pull/169

### DIFF
--- a/howto/customize_paths.md
+++ b/howto/customize_paths.md
@@ -1,0 +1,33 @@
+## Customize paths
+
+The project is setup to use specific paths to read the original weights and save checkpoints etc.
+
+For all scripts, you can run
+
+```shell
+python script.py -h
+```
+
+to get a list of available options. For instance, here's how you would modify the checkpoint dir:
+
+```shell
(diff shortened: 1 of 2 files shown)


lantiga (commented):
Looks good, although it might be hard to find if you're in the weeds.
I suggest adding links to this from the other howtos.

carmocca:
Done @lantiga! example: https://github.com/Lightning-AI/lit-llama/blob/carmocca/howto-customize/howto/inference.md

howto/customize_paths.md
```diff
@@ -0,0 +1,33 @@
...
+For all scripts, you can run
+
+```shell
+python script.py -h
+```
+
+to get a list of available options. For instance, here's how you would modify the checkpoint dir:
+
+```shell
+python scripts/convert_checkpoint.py --checkpoint_dir "/srv/data/checkpoints/foo"
```
awaelchli, line 14: if we could remove the /srv prefix here, I think it looks a bit odd :))
Just `/data/...` would be enough

awaelchli (approved):
I would have preferred to inline this info into the individual instruction files, but I'm fine with this too :) In any case, the info is useful to have, thanks!

PR #173, opened 2023-04-19
Adding a quantized kernel through triton

t-vi:
Performance is still far from where we want (4.4 tokens/s compared to 12.8 without quantization when generating 512 new tokens).
Future improvements could be:
- try to move to int32 to contain the quantized weights. This might have aligned reads more.
- try to be more efficient about reading these multiple times, but I did get bad performance when trying to put that into triton.

There also is quite a numerical difference in the results, apparently from the triton.dot (maybe it uses tf32?). Unfortunately triton segfaulted when I tried to verify with float64.


awaelchli (approved):

lantiga (approved):
Looks great. And advanced :-)

PR #174, opened 2023-04-20
Real ROPE

t-vi:
Here is a non-complex-using implementation of the ROPE.
In my experiments, casting the cache to float16 (not leaving it in bfloat16 or some crazyness) has been key to getting identical behaviour.
Because this is finicky, I have left the old implementation in and switcheable by setting the environment variable
`__LLAMA_REFERENCE_ROPE=1`

The idea here is that if everyone is happy with the new, we can drop the old implementation.


t-vi:
@awaelchli Could you take a look if old vs. new is close enough?

t-vi:
Oops, this was accidentally including the quantization, dropped that by rebasing + force push.

tests/test_rope.py
```diff
@@ -57,12 +57,11 @@ def test_rope(lit_llama, orig_llama) -> None:
...
+    torch.testing.assert_close(freqs_cis, torch.view_as_complex(llama_rope_cache))
 
     llama_x_rope = lit_llama.apply_rope(x.transpose(1, 2), llama_rope_cache).transpose(1, 2)
     orig_llama_x_rope, _ = orig_llama.apply_rotary_emb(x, x, freqs_cis)
-
-    assert torch.equal(llama_x_rope, orig_llama_x_rope)
+    torch.testing.assert_close(llama_x_rope, orig_llama_x_rope)
 
     # For posterity, we show here that our older implementation we initially wanted to use
     # is not numerically equivalent to Meta's rope implementation
```
carmocca, line 67: The old implementation in this file, this comment and the subsequent check can now be removed

lantiga (approved):
This looks awesome @t-vi!

lantiga:
Let's wait for @awaelchli's check, since he spent some time in the past on this issue

carmocca:
Can you comment here the output of some generations for the same prompt using each implementation?

awaelchli:
I generated several samples with generate.py on the main branch vs. this PR, and the outputs are identical!

lit_llama/model.py
```diff
@@ -231,3 +231,50 @@ def apply_rope(x: torch.Tensor, rope_cache: torch.Tensor) -> torch.Tensor:
...
+    x_out2 = torch.stack(
+        [xshaped[..., 0] * rope_cache[..., 0] - xshaped[..., 1] * rope_cache[..., 1],
+         xshaped[..., 1] * rope_cache[..., 0] + xshaped[..., 0] * rope_cache[..., 1],
+        ], -1)
+
+    x_out2 = x_out2.flatten(3)
+    return x_out2.transpose(1, 2).type_as(x)
+
+if os.environ.get('__LLAMA_REFERENCE_ROPE', '')[:1] in {'Y', 'y' ,'1'}:
+    print("USING REFERENCE ROPE")
```
awaelchli, line 278: I would be fine with completely dropping the old one.
And if we keep it, I vote for moving the old out of this file.
@lantiga ?

awaelchli (approved):

lit_llama/model.py
lantiga: Agreed, let's just drop it

t-vi: Yeah, it was only for easy of comparison. :)

tests/test_rope.py
t-vi: OK!

PR #178, opened 2023-04-20
Compute the generation time per sample

carmocca:
With compilation-based strategies or using `torch.compile`, the generation time might vary a lot across generations. The current implementation computes the total duration which is not that useful in this context.

This PR proposes to print the generation time per generation.

Right now, it interleaves the generation with the time. If you prefer, I can save either of them into a list and flush them together at the end.

I can update the other generation scripts once the design is decided.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -126,9 +126,8 @@ def main(
     encoded_prompt = tokenizer.encode(prompt, bos=True, eos=False, device=fabric.device)
 
     L.seed_everything(1234)
-    t0 = time.perf_counter()
-
-    for _ in range(num_samples):
+    for i in range(num_samples):
+        t0 = time.perf_counter()
         y = generate(
             model,
             encoded_prompt,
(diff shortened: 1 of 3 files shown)


generate.py
```diff
@@ -137,10 +136,9 @@ def main(
             temperature=temperature,
             top_k=top_k,
         )
+        t = time.perf_counter() - t0
         print(tokenizer.decode(y))
-
-    t = time.perf_counter() - t0
-    print(f"\n\nTime for inference: {t:.02f} sec total, {num_samples * max_new_tokens / t:.02f} tokens/sec", file=sys.stderr)
+        print(f"Time for inference {i + 1}: {t:.02f} sec total, {max_new_tokens / t:.02f} tokens/sec", file=sys.stderr)
```
awaelchli, line 141: What if we just drop the total time and print the tok/sec for the last iteration of the loop (unindenting the print)? The total time is not that useful tbh. The averaging over just a single generate() call should be good enough as estimate for tok/sec.

Also, the changes would have to be made for the other generate_* scripts.

carmocca: I personally like having the extra information for each generation. Anyways most people will run `num_samples=1` so there's no difference in that case.

lantiga (approved):
Looks good

generate_adapter.py
```diff
@@ -95,16 +93,15 @@ def main(
         top_k=top_k,
         eos_id=tokenizer.eos_id
     )
+    t = time.perf_counter() - t0
 
     output = tokenizer.decode(output)
     output = output.split("### Response:")[1].strip()
-
     print(output)
```
carmocca, line 100: @awaelchli did you intend to include the decoding + splitting + print inside the timer?

awaelchli: Yes, I think it should be, because that's the end result the user is waiting for in the end. print and str.split will be negligible.

PR #103, opened 2023-04-06
WIP: Make LLaMA `torch.compile` compatible

awaelchli:
Builds on top of #100
Attempt to address #62 by introducing a switch between the complex and non-complex variant of the rope implementation:
- complex: used to be numerically in parity with Meta's model and checkpoints (inference)
- non-complex: Used for training and finetuning

**Drawback:** The switch makes the code less readable and introduces indirection. The reader may struggle understanding the nuance here. If this is going too much against our minimalistic principles, we should drop this idea. Suggestions welcome!

**Adapter**
no compile + dynamic padding: between 90ms and 160ms
no compile + full padding: 352 ms
torch.compile + full padding: 118 ms

PR #179, opened 2023-04-20
Add simple test asserting LLaMA can torch.compile

awaelchli:
Adds a simple test now that LLaMA can `torch.compile`. 
Scripts will be modified to use torch.compile in #103

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -177,3 +177,20 @@ def test_adapter_parity(orig_llama_adapter):
     expected = orig_llama_model(token_sample, 0)
     out = llama_model(token_sample)
     assert torch.allclose(out, expected)
+
+
+def test_model_compile(lit_llama):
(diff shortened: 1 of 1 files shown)


carmocca (approved):

tests/test_model.py
```diff
@@ -176,4 +178,23 @@ def test_adapter_parity(orig_llama_adapter):
 
     expected = orig_llama_model(token_sample, 0)
     out = llama_model(token_sample)
-    assert torch.allclose(out, expected)
\ No newline at end of file
+    assert torch.allclose(out, expected)
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="torch.compile not supported on Windows")
```
carmocca, line 184:
```suggestion
@pytest.mark.skipif(sys.platform in ("win32", "darwin"), reason="torch.compile not supported on this platform")
```

lantiga (approved):

Issue #182, opened 2023-04-21
Incorrect DATA_FILE_NAME  in prepare_alpaca.py script


lucas-ventura:
Description: The prepare_alpaca.py script in the repository contains an error in the definition of the DATA_FILE_NAME variable. The current value is "[alpaca_data_cleaned.json](https://github.com/Lightning-AI/lit-llama/blob/main/scripts/prepare_alpaca.py#L17)", while it should be "alpaca_data_cleaned_archive.json". 

Suggested Solution:
Update the value of the DATA_FILE_NAME variable to "alpaca_data_cleaned_archive.json" in the prepare_alpaca.py script to match the actual filename.
```
DATA_FILE = "https://raw.githubusercontent.com/tloen/alpaca-lora/main/alpaca_data_cleaned_archive.json"
DATA_FILE_NAME = "alpaca_data_cleaned_archive.json"
```

awaelchli:
They have previously renamed it from alpaca_data_cleaned to alpaca_data_cleaned_archive. For us the only thing that matters is that the URL is correct, since our scripts don't need to directly load this file but our preprocessed ones. Feel free to send a PR correcting the inconsistency if you are interested :)

PR #183, opened 2023-04-21
Fixes DATA_FILE_NAME with updated from DATA_FILE

lucas-ventura:
This pull request addresses issue #182 by updating the value of the DATA_FILE_NAME variable in the prepare_alpaca.py script to "alpaca_data_cleaned_archive.json", which is the correct file name. The change is a simple modification to a single line of code.


lantiga (commented):
Looks good, thank you @lucas-ventura

lantiga (approved):

carmocca:
Just for reference, this was left as it was on purpose when the URL changed: https://github.com/Lightning-AI/lit-llama/pull/136#discussion_r1165991021. Either way, it's okay to change it too

PR #185, opened 2023-04-21
Fix .eval() merging of LoRA weights

awaelchli:
When we adapted the LoRA layer implementation from https://github.com/microsoft/LoRA/ we also inherited a bug in the train() and eval() methods that are meant to merge and unmerge the LoRA weights.

This was already pointed out by users on their side https://github.com/microsoft/LoRA/issues/34 but was never addressed.
The problem is that they implement the eval() method, which is dead code and never gets called. Because internally in `nn.Module.eval()`, it calls `nn.Module.train(False)` and thus only ever the train() method gets called recursively on all submodules.

### DIFF
--- a/lit_llama/lora.py
+++ b/lit_llama/lora.py
@@ -86,14 +86,8 @@ class MergedLinear(nn.Linear, LoRALayer):
 
     def zero_pad(self, x):
         x = x.transpose(0, 1)
-        # print("x", x.shape) # 64, 32
-        # print("lora_ind", self.lora_ind)  # [32 * True, 32 * False, 32 * True]
-        # print("out_features", self.out_features)
         result = x.new_zeros((*x.shape[:-1], self.out_features))
-        # print("result 1", result.shape)
         result = result.view(-1, self.out_features)
-        # print("result 2", result.shape)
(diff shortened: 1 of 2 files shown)


lit_llama/lora.py
```diff
@@ -84,42 +85,34 @@ def reset_parameters(self):
...
 
     def zero_pad(self, x):
+        x = x.transpose(0, 1)
         result = x.new_zeros((*x.shape[:-1], self.out_features))
         result = result.view(-1, self.out_features)
         result[:, self.lora_ind] = x.reshape(
             -1, self.out_features // len(self.enable_lora) * sum(self.enable_lora)
         )
-        return result.view((*x.shape[:-1], self.out_features))
+        return result.view((*x.shape[:-1], self.out_features)).transpose(0, 1)
```
awaelchli, line 94: this bug was hidden behind train() and eval() never triggering the actual merge/unmerge. We would get shape mismatches without the transpose. 🤷

carmocca (approved):
Good catch!

lantiga (approved):
🎩 hat tip @awaelchli

PR #190, opened 2023-04-24
Reference Lit-LLaMA articles and tutorials

rasbt:
Cross-links some additional Lit-LLaMA resources we have.

Not sure if that's too much though. Let's discuss I'd say!

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ Look at `train.py` for a starting point towards pre-training / fine-tuning using
 
 We welcome all individual contributors, regardless of their level of experience or hardware. Your contributions are valuable, and we are excited to see what you can accomplish in this collaborative and supportive environment. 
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -125,11 +125,21 @@ It is expected that you have downloaded the pretrained weights as described abov
...
+These technical tutorials illustrate how to run the finetuning code.
 
 - [Finetune with LoRA](howto/finetune_lora.md)
 - [Finetune with Adapters](howto/finetune_adapter.md)
 
+### Understanding Finetuning -- Conceptual Tutorials
+
+Looking for conceptual tutorials and explanations? We have some additional articles below:
+
+- [Understanding Parameter-Efficient Finetuning of Large Language Models: From Prefix Tuning to LLaMA-Adapters](https://lightning.ai/pages/community/article/understanding-llama-adapters/)
```
awaelchli, line 141: I think it would be good to include this reference inside `howto/finetune_adapter.md` as well (somewhere near the introductory text perhaps?). I thought I did that already but it must have gotten lost somehow. 

And of course we would love to include the LoRA article as well when it comes out.

rasbt: Good call! Added that. (And I made a note to add the LoRA article as well later this week after that has gone live!)

lantiga (approved):
Looks good!

PR #192, opened 2023-04-24
Add packed dataset

lantiga:
~Combined dataset testing is still TODO, the rest is ready for review.~

CombinedDataset will come in a later PR

### DIFF
--- a/lit_llama/packed_dataset.py
+++ b/lit_llama/packed_dataset.py
@@ -4,7 +4,6 @@
 
 import os
 import struct
-import random
 
 import numpy as np
 from torch.utils.data import IterableDataset, get_worker_info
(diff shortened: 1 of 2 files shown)


lit_llama/packed_dataset.py
```diff
@@ -0,0 +1,237 @@
...
+            else range(n_all_blocks)
+        )
+
+        self._curr_idx = 0
+
+    # def __getstate__(self):
+    #     # TODO: update
+    #     return self._filenames, self._n_chunks, self._seed
+
+    # def __setstate__(self, state):
```
awaelchli, line 111: When would these be invoked?

awaelchli (approved):

lit_llama/packed_dataset.py
awaelchli: Was there some pickling going on?

lit_llama/packed_dataset.py
```diff
@@ -0,0 +1,237 @@
...
+        elem_id = (block_idx % self._n_blocks) * self._block_size
+        offset = np.dtype(self._dtype).itemsize * elem_id
+        arr = np.frombuffer(
+            buffer, dtype=self._dtype, count=self._block_size, offset=offset
+        )
+        self._curr_idx += 1
+        return arr
+
+
+class PackedDataset(IterableDataset):
```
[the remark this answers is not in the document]
lantiga: Yes you're right

lit_llama/packed_dataset.py
lantiga: Eventually being able to serialize and deserialize the dataset deterministically is going to be important but I removed for now.

PR #194, opened 2023-04-26
Update default values for max sequence length

awaelchli:
Block size caused some confusion:
1. The default in the model was set to 4096, but the [reference code from meta uses 2048](https://github.com/facebookresearch/llama/blob/57b0eb62de0636e75af471e49e2f1862d908d9d8/llama/model.py#L30).
2. We reduce the max sequence length (referred to block_size because we use the nano-gpt terminology) because we train on shorter sequences from the alpaca dataset. 

The PR adjusts these values, making them consistend between the finetuning scripts.
The changes have no impact on the training, because all sequences are always shorter than the values adjusted here.


lit_llama/model.py
```diff
@@ -13,7 +13,7 @@
 
 @dataclass
 class LLaMAConfig:
-    block_size: int = 4096
+    block_size: int = 2048
```
awaelchli, line 16: as in https://github.com/facebookresearch/llama/blob/57b0eb62de0636e75af471e49e2f1862d908d9d8/llama/model.py#L30

lantiga (approved):
Good call

PR #199, opened 2023-04-27
Make Red Pajama prepare scripts use PackedDataset

lantiga:
Follow-up to #192 
The prepare_redpajama scripts for sample and full now use the new packed dataset format. Tests are included.

Since Red Pajama relies on a mix of dataset, this PR also introduces a way to read the mix with predetermined mixing weights.


scripts/prepare_redpajama.py
```diff
@@ -44,6 +44,7 @@ def prepare_sample(
     source_path: Path,
     tokenizer_path: Path,
     destination_path: Path,
+    chunk_size: int,
```
awaelchli, line 47: Could you remind me what the chunk size is? It is only relevant for writing, right? It is unrelated to how the data is preprocessed and packed, correct?

awaelchli (approved):

scripts/prepare_redpajama.py
lantiga: exactly, it's the size of each file being written, expressed as the number of tokens
if you use a chunk_size that is a multiple of block_size you'll use when you read, then you'll get all the tokens in a chunk, otherwise you might cut a few off at the end of each chunk, but you'll still be able to read
it shouldn't be too much of an issue on large datasets

PR #202, opened 2023-04-27
evaluate_*.py scripts for lora and adapter

rasbt:
Technically, we can skip the lines

```python
sample = {"instruction": test_string, "input": input}
test_string = generate_prompt(sample)
```

but the modeling performance is the same either way. (On some benchmarks, it's like 0.01 perplexity points worse but negligible.)

Let me know your preference @lantiga @awaelchli @carmocca 

And otherwise the PR should be ready.

### DIFF
--- a/evaluate_adapter.py
+++ b/evaluate_adapter.py
@@ -1,0 +1,164 @@
+# This mimics GPTQ's evaluation metrics: https://github.com/IST-DASLab/gptq/
+# Thanks to E. Frantar et al GPTQ: Accurate Post-training Compression for GPT, arXiv:2210.17323
+import math
+import sys
+import time
+from pathlib import Path
+from typing import Optional
+
+import lightning as L
+import torch
+import tqdm
+
+from lit_llama import Tokenizer
+from lit_llama.adapter import LLaMA
+from lit_llama.utils import EmptyInitOnDevice, lazy_load, llama_model_lookup
+from scripts.prepare_alpaca import generate_prompt
+
+from datasets import load_dataset
+
+
+def load_eval_data(dataset_name: str) -> str:
+    # this mimics gptq datautils
+    if dataset_name == "wikitext":
+        # traindata = load_dataset('wikitext', 'wikitext-2-raw-v1', split='train')
+        testdata = load_dataset("wikitext", "wikitext-2-raw-v1", split="test")
+        testdata = "\n\n".join(testdata["text"])
+    elif dataset_name == "ptb":
+        testdata = load_dataset("ptb_text_only", "penn_treebank", split="test")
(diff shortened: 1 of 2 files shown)


evaluate_adapter.py
```diff
@@ -0,0 +1,164 @@
...
+    total_toks = 0
+    model = fabric.setup_module(model)
+
+    tokenizer = Tokenizer(tokenizer_path)
+
+    for dsname in datasets.split(","):
+        test_string = load_eval_data(dsname)
+
+        sample = {"instruction": test_string, "input": input}
+        test_string = generate_prompt(sample)
```
awaelchli, line 120: @rasbt In the PR description, you wrote that these lines could be removed. Do you mean that the encoding into the instruction template is not necessary, and the prompt can be given as is? But how would the instructiono be combined with the optional "input"?

awaelchli (approved):

evaluate_adapter.py
rasbt: @awaelchli 

Yeah, so what I meant was instead of

```python
       test_string = load_eval_data(dsname)

         sample = {"instruction": test_string, "input": input}
         test_string = generate_prompt(sample)

         encoded_text = tokenizer.encode(
             test_string, bos=True, eos=False, device=fabric.device
         )
```

we can simply do

```python
       test_string = load_eval_data(dsname)
       encoded_text = tokenizer.encode(
             test_string, bos=True, eos=False, device=fabric.device
       )
```

Since we are only evaluating perplexity, I think it doesn't matter so much here. We are basically just checking if the model can generate "likely" text. So, in other words, it's not really evaluating the instruction-output pair yet. That would be some future todo item for more comprehensive Q&A evaluation.

So maybe we should leave it as

```python
       test_string = load_eval_data(dsname)

         sample = {"instruction": test_string, "input": input}
         test_string = generate_prompt(sample)

         encoded_text = tokenizer.encode(
             test_string, bos=True, eos=False, device=fabric.device
         )
```

to check whether the model can handle this format though?

awaelchli: Okay, sounds good. If you remove it, could you add a small comment in that line that the conversion to the template is missing and why? Thanks!

rasbt: Oh I would maybe keep it because it's the "more complete" code. What do you think?

awaelchli: Yep, good with me. Good to merge this then?

rasbt: Yes would be good to merge. (Btw I compared the two approaches mentioned above, and the performance was basically the same. In some cases there was a tiny 0.01 point difference which doesn't really matter in practice)

PR #209, opened 2023-05-01
WIP: Pre-training script

lantiga:


train_redpajama.py
```diff
@@ -0,0 +1,297 @@
...
+        model = LLaMA(config).bfloat16()
+        model.apply(model._init_weights)
+        torch.set_default_tensor_type(torch.FloatTensor)
+
+    # if compile:
+    #     model = torch.compile(model)
+
+    model = fabric.setup_module(model)
+
+    train_dataloader, val_dataloader = create_dataloaders(
```
awaelchli, line 92: Since we are using torch dataloaders here, we could call `fabric.setup_dataloaders()` as well. This would properly add the distributed sampler and then we can/should use the same seed on every rank.

train_redpajama.py
```diff
@@ -0,0 +1,297 @@
...
+    )
+    strategy = FSDPStrategy(
+        auto_wrap_policy=auto_wrap_policy, activation_checkpointing=Block
+    )
+
+    fabric = L.Fabric(
+        accelerator="cuda", devices=devices, precision="bf16-mixed", strategy=strategy
+    )
+    fabric.launch()
+    fabric.seed_everything(1337 + fabric.global_rank)
```
awaelchli, line 74:
```suggestion
    fabric.seed_everything(1337)
```
Using the same seed for all ranks would be better in this case here (see other comment).

awaelchli (approved):

train_redpajama.py
lantiga: You're 100% right

PR #210, opened 2023-05-01
Add missing save_model_checkpoint import

lucas-ventura:
## Changes Made

I added the following line to the `finetune_full.py` file:

```python
from lit_llama.utils import save_model_checkpoint
```


lantiga (approved):
Thank you @lucas-ventura !

PR #213, opened 2023-05-02
Set up dataloaders in redpajama training script

awaelchli:
`Fabric.setup_dataloaders()` takes care of attaching a distributed sampler, which was missing here. Doing so will also move the data automatically to the device so we can delete the manual `fabric.to_device()`

### DIFF
--- a/train_redpajama.py
+++ b/train_redpajama.py
@@ -135,10 +135,9 @@ def train(
 
         t0 = time.time()
 
-        input_ids, targets = get_batch(
-            fabric, train_data, block_size=model.config.block_size
-        )
-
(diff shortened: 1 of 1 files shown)


train_redpajama.py
```diff
@@ -266,9 +267,8 @@ def create_dataloaders(
 def get_batch(
```
awaelchli, line 267: now that there are only two lines in this function, should I inline that code in our training loop?

lantiga: yes let's do it

lantiga (approved):
Looks great, thank you @awaelchli

train_redpajama.py
awaelchli: done

PR #214, opened 2023-05-02
Fix copy path for tokenizer model when converting meta checkpoint

### LINKED ISSUE #206: [BUG] Cannot run scripts/convert_checkpoint.py anymore
https://github.com/Lightning-AI/lit-llama/blob/main/howto/download_weights.md

Issue: The scripts/convert_checkpoint.py does not work anymore due to wrong directory has been passed.

Solution:
In convert_checkpoint.py line 66 to 79, 
```
def meta_weights_for_nano_model(
    *,
    output_dir: Path = Path("checkpoints/lit-llama"),
    ckpt_dir: Path = Path("checkpoints/llama/"),
    llama_root_dir = Path("checkpoints/llama/"),
    model_size: str = "7B",
    dtype: str = "float32",
) -> None:
    output_dir = output_dir / model_size
    ckpt_dir = ckpt_dir / model_size
    output_dir.mkdir(parents=True, exist_ok=True)

    # the tokenizer is the same for all model sizes, so we store it in the parent dir
    shutil.copy(llama_root_dir / "tokenizer.model", output_dir.parent)
```

or just use ```shutil.copy(ckpt_dir.parent / "tokenizer.model", output_dir.parent)```

Please fix and thank you.

awaelchli:
Fixes #206

### DIFF
--- a/scripts/convert_checkpoint.py
+++ b/scripts/convert_checkpoint.py
@@ -75,7 +75,7 @@ def meta_weights_for_nano_model(
     output_dir.mkdir(parents=True, exist_ok=True)
 
     # the tokenizer is the same for all model sizes, so we store it in the parent dir
-    shutil.copy(ckpt_dir / "tokenizer.model", output_dir.parent)
+    shutil.copy(ckpt_dir.parent / "tokenizer.model", output_dir.parent)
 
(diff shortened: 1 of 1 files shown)


lantiga (approved):
LGTM

PR #217, opened 2023-05-03
Add tokens/sec/device and new Fabric initialization

lantiga:
Getting 2k tokens/sec/GPU on RedPajama with this configuration (FSDP + activation checkpointing) on 4xA100 40GB

Note that:
- we are not using torch.compile since activation checkpointing doesn't seem to be supported as of today
- the original LLaMA implementation uses fused kernels from xformers

### DIFF
--- a/train_redpajama.py
+++ b/train_redpajama.py
@@ -88,10 +88,10 @@ def main(
     train_dataloader, val_dataloader = fabric.setup_dataloaders(train_dataloader, val_dataloader)
 
     with fabric.device:
-        torch.set_default_tensor_type(torch.HalfTensor)
-        model = LLaMA(config).bfloat16()
+        torch.set_default_dtype(torch.bfloat16)
+        model = LLaMA(config)
         model.apply(model._init_weights)
-        torch.set_default_tensor_type(torch.FloatTensor)
+        torch.set_default_dtype(torch.float32)
 
     # if compile:
     #     model = torch.compile(model)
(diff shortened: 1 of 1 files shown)


train_redpajama.py
```diff
@@ -98,18 +87,28 @@ def main(
     )
     train_dataloader, val_dataloader = fabric.setup_dataloaders(train_dataloader, val_dataloader)
 
+    with fabric.device:
+        torch.set_default_tensor_type(torch.HalfTensor)
```
justusschock, line 91: maybe change this to [`torch.set_default_dtype`](https://pytorch.org/docs/stable/generated/torch.set_default_dtype.html#torch.set_default_dtype) as this method will be deprecated (https://github.com/pytorch/pytorch/pull/99521)

train_redpajama.py
```diff
@@ -98,18 +87,28 @@ def main(
...
+        model = LLaMA(config).bfloat16()
+        model.apply(model._init_weights)
+        torch.set_default_tensor_type(torch.FloatTensor)
```
justusschock, line 94: same as above :)

train_redpajama.py
```diff
@@ -119,6 +118,7 @@ def train(
     train_dataloader: DataLoader,
     val_dataloader: Optional[DataLoader],
     grad_accum_steps: int,
+    devices: int,
```
justusschock, line 121: this seems to be unused :)

justusschock (approved):

awaelchli (approved):

train_redpajama.py
lantiga: Good catch, you're right. It's a remnant of when I was using it to compute token / sec / GPU in a different way

train_redpajama.py
lantiga: Yep

PR #218, opened 2023-05-03
Implement data-sharding in PackedDataset

awaelchli:
Follow-up to #217

The iterable dataset implementation for redpajama so far supported only sharding the data across the dataloader workers, but not globally across all processes in a distributed setting, e.g., Fabric(devices > 1).

### DIFF
--- a/train_redpajama.py
+++ b/train_redpajama.py
@@ -78,17 +78,6 @@ def main(
 
     config = LLaMAConfig.from_name("7B")
 
-    with fabric.device:
-        torch.set_default_tensor_type(torch.HalfTensor)
-        model = LLaMA(config).bfloat16()
-        model.apply(model._init_weights)
-        torch.set_default_tensor_type(torch.FloatTensor)
-
-    # if compile:
-    #     model = torch.compile(model)
-
-    model = fabric.setup_module(model)
-
     train_dataloader, val_dataloader = create_dataloaders(
         batch_size=micro_batch_size,
         block_size=config.block_size,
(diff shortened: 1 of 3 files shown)


train_redpajama.py
```diff
@@ -226,14 +227,16 @@ def create_dataloader(
...
     shuffle: bool = True,
     seed: int = 12345,
 ) -> DataLoader:
     datasets = []
     for prefix, _ in data_config:
         filenames = glob.glob(os.path.join(data_dir, prefix + "*"))
         dataset = PackedDataset(
-            filenames, n_chunks=10, block_size=block_size, shuffle=shuffle, seed=seed
+            filenames, n_chunks=4, block_size=block_size, shuffle=shuffle, seed=seed,
+            num_processes=fabric.world_size, process_rank=fabric.global_rank,
```
awaelchli, line 239: the n_chunks value here must be smaller than the number of files ending up in each shard. It's unclear to me what exactly the chunks refer to

lantiga: Say you have 12 files (12 chunks, each chunk containing N blocks of tokens) and you set `n_chunks=4`.
The dataset will pre-load 4 chunks in memory and iterate over the blocks contained in those chunks, optionally shuffling the blocks. If the dataset doesn't find enough chunks to load, it will raise a StopIteration.

I think we should actually give the opportunity to rewind as well for overfit tests etc.

lantiga (approved):
Looks great

PR #220, opened 2023-05-03
 Refactor model saving logic to handle different strategies

lucas-ventura:
Updates the logic for saving model checkpoints to handle different strategies. Previously, the code was throwing a `ValueError` when trying to save a checkpoint using the `fabric.save()` method for a model trained with FSDP strategy. Now, the code uses the `save_model_checkpoint()` function to handle the saving of the checkpoint, which can handle different strategies.


awaelchli (approved):
I agree with this for an intermediate solution. Good catch!

FYI, this is a temporary fix for now, and we are currently working on making the saving/loading experience a bit smoother (by supporting the converstion of the different state dict types) so that all strategies can use the unified fabric.save/load api without going through this utility.

PR #222, opened 2023-05-03
Small model improvements

### LINKED ISSUE #221: Suggest multiple of 64 for vocab size during training and finetuning
https://twitter.com/karpathy/status/1621578354024677377

carmocca:
Fixes #221 

Changes:

- Remove `c_` prefix that stems from the fact that gpt2 used convolutional layers. Removed in https://github.com/karpathy/minGPT/commit/7e68832554096ddea9e178e8d80a6679491a6ff6
- Rename `rms` layers to `norm` so users can easily switch the norm type without changing the attribute name
- Rename `fc1` to `fc_1` for consistency with other numbered layers (e.g `rms_1`)
- Automatically find a padded vocab size.

This is a breaking change for existing trained lit-llama checkpoints

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -24,7 +24,7 @@ jobs:
           - {os: "macOS-11", python-version: "3.10"}
           - {os: "ubuntu-20.04", python-version: "3.10"}
           - {os: "windows-2022", python-version: "3.10"}
-    timeout-minutes: 10
+    timeout-minutes: 15
 
     steps:
     - uses: actions/checkout@v3


lit_llama/model.py
```diff
@@ -5,20 +5,29 @@
...
     block_size: int = 2048
     vocab_size: int = 32000
+    padded_vocab_size: Optional[int] = None
     n_layer: int = 32
     n_head: int = 32
     n_embd: int = 4096
 
+    def __post_init__(self):
+        if self.padded_vocab_size is None:
+            self.padded_vocab_size = find_multiple(self.vocab_size, 64)
```
carmocca, line 29: We could throw a warning here if we want to. I thought it was unnecessary though because the user would already be doing the conscious choice of passing `padded_vocab_size=non_64_multiple_vocab_size`

awaelchli:
This will invalidate all the checkpoints we have trained (and users) and they would have to be converted. Is it really worth the price for this cosmetic change? I don't see enough value in just changing a couple underscore prefixes, so I'm voting against doing it.

lantiga:
I agree with @awaelchli, I originally didn't realize it would break things for everyone (of course the other repo can have all these changes).

What can we keep that doesn't break compatibility?

carmocca:
Right, so this is why I asked if it was too late. I'll revert the breaking changes and just add the padding change

PR #223, opened 2023-05-03
Automatic vocabulary size padding

### LINKED ISSUE #221: Suggest multiple of 64 for vocab size during training and finetuning

carmocca:
Fixes #221

Part of #222

### DIFF
--- a/lit_llama/model.py
+++ b/lit_llama/model.py
@@ -5,19 +5,28 @@ Based on the nanoGPT implementation: https://github.com/karpathy/nanoGPT.
 # mypy: ignore-errors
 import math
 from dataclasses import dataclass
+from typing import Optional
 
 import torch
 import torch.nn as nn
 from torch.nn import functional as F
 from typing_extensions import Self
 
+from lit_llama.utils import find_multiple
+
+
 @dataclass
 class LLaMAConfig:
     block_size: int = 2048
     vocab_size: int = 32000
+    padded_vocab_size: Optional[int] = None
     n_layer: int = 32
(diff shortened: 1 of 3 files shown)


lit_llama/model.py
```diff
@@ -35,14 +44,13 @@ def from_name(cls, name: str) -> Self:
...
-        assert config.block_size is not None
+        assert config.padded_vocab_size is not None
         self.config = config
 
-        self.lm_head = nn.Linear(config.n_embd, config.vocab_size, bias=False)
+        self.lm_head = nn.Linear(config.n_embd, config.padded_vocab_size, bias=False)
         self.transformer = nn.ModuleDict(
             dict(
-                wte=nn.Embedding(config.vocab_size, config.n_embd),
+                wte=nn.Embedding(config.padded_vocab_size, config.n_embd),
```
awaelchli, line 53: The current default vocab size is a multiple of 64, so this should be save since the tensor size here won't actually be different to the shape in the checkpoint.

lit_llama/model.py
```diff
@@ -103,7 +111,7 @@ def __init__(self, config: LLaMAConfig) -> None:
         self.n_head = config.n_head
         self.n_embd = config.n_embd
         self.block_size = config.block_size
-        self.rope_cache = None
+        self.rope_cache: Optional[torch.Tensor] = None
```
awaelchli, line 114: I wouldn't add the typing 😅

awaelchli (approved):

PR #229, opened 2023-05-07
Fix dolly dataset download link

vihangd:
Fix dolly dataset link as databricks have migrated the dataset to HF


lantiga (approved):
Thank you @vihangd!

PR #238, opened 2023-05-08
Fix Gradient Accumulation Logic

waitzkin:
This fixes the gradient accumulation logic in finetune_full.py.

Since weight update should occur once in every {gradient_accumulation_steps} iterations, I think "==" in line 100 should be "!=".


lantiga (approved):
Great catch @waitzkin, we got this right in the pretraining script but not here.

PR #240, opened 2023-05-08
Add Cached KVs

gkroiz:
Helping to merge @Liyang90's changes to improve inference speed via cached KVs.

There are several key changes in these 5 commits I made:
- added xm.mark_step() in the token generation loop, so each token generation executes a graph.
- added key, value caching in both generation and model forward code. The forward now only need the new token and the cached kvs as inputs.
- moved rope cache to the whole model from each attention layer, because all the layers can share the same rope cache.
- added mask cache to use for attention masking with cached kvs.
- fix ops and tensors that would cause recompilation.


carmocca (commented):
Hey @gkroiz and @Liyang90! Thank you for this

It looks to me like this PR is mixing XLA improvements with the KV cache implementation.

Could we split this into two?

The KV cache implementation could be merged into the script, but for the XLA code we would probably duplicate it into a `model_xla`, so that we can specialize it as much as necessary without worrying about other accelerators.

Thank you!

gkroiz:
> Hey @gkroiz and @Liyang90! Thank you for this
> 
> It looks to me like this PR is mixing XLA improvements with the KV cache implementation.
> 
> Could we split this into two?
> 
> The KV cache implementation could be merged into the script, but for the XLA code we would probably duplicate it into a `model_xla`, so that we can specialize it as much as necessary without worrying about other accelerators.
> 
> Thank you!

Yes, we can split this PR into two separate PRs. If we create an `model_xla.py` file, would we also want to create an `xla` version of the training, finetuning, & other files?

carmocca:
It probably depends on the extent of the changes. But I would suggest to start off with a copy since we can always merge them later if it makes sense

PR #242, opened 2023-05-08
Fix checkpoint name

Ever2after:
Checkpoint name in generate_lora.py 'adapter_checkpoint' seems better to be changed to 'lora_checkpoint'.


awaelchli:
I think this is a good change. Let's do it?

lantiga (approved):
Yep! Thank you @Ever2after

PR #250, opened 2023-05-09
bug resolve in scripts/convert_hf_checkpoint.py

avinashreddych:
check_valid_checkpoint_dir(checkpoint_dir) before conversion of weights causes and error as the files doesn't exist before convert.


carmocca:
Hi! Good catch. Your change is good, but this is pointing to the wrong repository. Re-opened it for you at https://github.com/Lightning-AI/lit-parrot/pull/41

PR #262, opened 2023-05-11
Clarify how to use int4 quantization

rasbt:
It was not quite clear how to use the int4 quantized checkpoints after the conversion (and whether to use the dtype bfloat16 for this, which was referenced in the text). I added an example to clarify this.


lantiga (approved):
Awesome! Thank you @rasbt

Issue #197, opened 2023-04-27
KV cache for faster generation


carmocca:
Reference: https://www.dipkumar.dev/becoming-the-unbeatable/posts/gpt-kvcache/
Whisper (MIT licensed) has a nice PyTorch-y implementation for this: https://github.com/openai/whisper/blob/c09a7ae299c4c34c5839a76380ae407e7d785914/whisper/model.py#L274-L305
nanoGPT doesn't implement this.

An argument against this is that it complicates the model code during training

Suggested by @Liyang90

PR #266, opened 2023-05-13
Adding cached KVs

### LINKED ISSUE #309: when will kv caching landing?
currently inference is so slow, much more slow than huggingface's transformer implementation, any plan on speed up?

gkroiz:
This PR adds the option to cache KVs for inference. 

Cached KVs portion of splitting https://github.com/Lightning-AI/lit-llama/pull/240 into 2 PRs

main changes:

added KV caching
moved rope cache to the whole model from each attention layer
added mask cache to use for attention masking with cached kvs.
fix ops and tensors that would cause recompilation.

Similar to https://github.com/Lightning-AI/lit-parrot/pull/51
Fixes https://github.com/Lightning-AI/lit-llama/issues/197
Fixes #309

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -17,10 +17,11 @@ from lit_llama.utils import EmptyInitOnDevice, lazy_load, llama_model_lookup
 
 @torch.no_grad()
 def generate(
-    model: torch.nn.Module,
+    model: LLaMA,
     idx: torch.Tensor,
     max_new_tokens: int,
-    max_seq_length: int,
+    *,
+    max_seq_length: Optional[int] = None,
     temperature: float = 1.0,
     top_k: Optional[int] = None,
     eos_id: Optional[int] = None,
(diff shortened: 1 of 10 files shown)


gkroiz:
TODO: update tests

lantiga:
For visibility: updates to lit-parrot caching
https://github.com/Lightning-AI/lit-parrot/pull/60

generate.py
```diff
@@ -17,10 +17,11 @@
 
 @torch.no_grad()
 def generate(
```
carmocca, line 19: The other scripts that call this function also need an update

carmocca (commented):
Implementation looks good!

generate.py
gkroiz: I'll try to get to this later today

carmocca: I fixed adapter. Need to do a check across the repo to see what's outdated.

Can you update the inference numbers in the tpu howto?

gkroiz:
@carmocca what is the difference between `generate.py` and `generate/full.py`?

carmocca:
@gkroiz It's basically a duplicate. See the description here https://github.com/Lightning-AI/lit-llama/pull/255

carmocca:
I opened https://github.com/Lightning-AI/lit-llama/issues/313 to track it. I wouldn't try to merge them here.

generate/full.py
```diff
@@ -79,7 +25,7 @@ def main(
     top_k: int = 200,
     temperature: float = 0.8,
     checkpoint_path: Optional[Path] = None,
-    tokenizer_path: Path = Path("checkpoints/lit-llama/tokenizer.model"),
+    tokenizer_path: Path = Path("../checkpoints/lit-llama/tokenizer.model"),
```
carmocca, line 28: The assumption from the howtos is that these scripts are called from the repo root, so `..` shouldn't be used

generate.py
gkroiz: Waiting to update these numbers for https://github.com/Lightning-AI/lit-parrot/pull/83#issuecomment-1558042678

carmocca (approved):
I think we're done here! The howto update can be done separately

lantiga (approved):
Awesome job!

PR #273, opened 2023-05-16
Fix typo in download_weights.md

ezyang:


lantiga:
Yikes, sorry about that @ezyang 

The intention was to change all `ckpt_dir` to `checkpoint_dir` and `download_weights.py` somehow escaped the treatment. Fixed in #274 

I'm closing this one (we want to fix the script and keep the howto as is).

PR #277, opened 2023-05-16
LoRA annotated

Andrei-Aksionov:
The goal of this PR is to make it easier for anyone to understand how the code for LoRA works by:
- providing input/output shapes of tensor
- reasoning behind each and every method
- putting LoRA schema at the top of the file so it's easier to see on a high level what is going on when confusion starts kicking in

Some annotations might seem redundant (because they are 😄 ) but I think it's better to have more info than less.
Nevertheless my goal is to create clear annotations/code comments, so if anyone sees how I can improve this PR - please let me now.

P.S. English is my second language, so when I tried to create concise comments they might have come a bit coarse. That means that any help in rephrasing is highly welcomed.

### DIFF
--- a/lit_llama/lora.py
+++ b/lit_llama/lora.py
@@ -215,7 +215,8 @@ class MergedLinear(nn.Linear, LoRALayer):
     def zero_pad(self, x):
         """Properly pad weight updates with zeros.
 
-        If we want to finetune queires and values, but not keys, then the weights update should be:
+        If, based on `self.enable_lora`, we want to finetune queires and values, but not keys,
+        then the weights update should be:
 
         [[ΔW,ΔW,ΔW, ..., 0,0,0, ..., ΔW,ΔW,ΔW,],
          [....................................],
@@ -224,6 +225,11 @@ class MergedLinear(nn.Linear, LoRALayer):
         ________________________________________
         | query         | key       | value    |
         ----------------------------------------
+
+        Parameters
+        ----------
+        x : torch.Tensor
+            tensor with weights update that will be padded with zeros if necessary
 
         Returns
         -------
--- a/lit_llama/lora.py
+++ b/lit_llama/lora.py
@@ -31,11 +31,11 @@ r"""
              ┆         x         ┆
              └───────────────────┘
 
-With LoRA (Low Ranking Adaptation) instead of learning weights of size d*d, we can freeze the pretrained weights and
-instead learn two matrices of size d*r and r*d (they will store weight updates for the pretrained weights): the
-number of parameters in this case will be reduced drastically (depending on the rank of course) yet after
-multiplication of matrices d*r and r*d we will get a matrix d*d which we can sum with frozen pretrained weights and
-thus finetune model.
+With LoRA (Low Ranking Adaptation: https://arxiv.org/abs/2106.09685) instead of learning weights of size d*d,
+we can freeze the pretrained weights and instead learn two matrices of size d*r and r*d (they will store weight updates
+for the pretrained weights): the number of parameters in this case will be reduced drastically (depending on the rank of
+course) yet after multiplication of matrices d*r and r*d we will get a matrix d*d which we can sum with frozen
+pretrained weights and thus finetune the model.
 
 The goal of this approach is to move weight updates into a separete matrix which is decomposed with
 two matrices of a lower rank.
@@ -172,7 +172,7 @@ class MergedLinear(nn.Linear, LoRALayer):
             self.weight.requires_grad = False # (384, 128)
 
             # Compute the indices
-            # Indices are needed to properly pad weight updates with zeros. If we want to finetune queires and values,
+            # Indices are needed to properly pad weight updates with zeros. If we want to finetune queries and values,
             # but not keys, then the weights update should be:
             #
             # [[ΔW,ΔW,ΔW, ..., 0,0,0, ..., ΔW,ΔW,ΔW,],
--- a/lit_llama/lora.py
+++ b/lit_llama/lora.py
@@ -7,27 +7,29 @@
 r"""
     Low Ranking Adaptation for LLMs scheme.
 
-            -----------------
-            |       h       |
-            -----------------
-                    ^
-                    |
-                    +
-                 /     \
-    ----------------    -----------          Matrix initialization:
-    |  pretrained  |     \   B    /          B = 0
-    |   weights    |      \ r*d  /           A = N(0, sigma^2)
-    |              |       -----
-    |  W e R^(d*d) |       | r |             r - rank
-    |              |       -----
-    ----------------      /  A  \
-            ^            /  d*r  \
-             \          -----------
-              \         ^
-               \       /
-            -----------------
-            |       x       |
-            -----------------
+             ┌───────────────────┐
+             ┆         h         ┆
+             └───────────────────┘
+                       ▲
+                       |
+                       +
+                    /     \
+    ┌─────────────────┐    ╭───────────────╮     Matrix initialization:
+    ┆                 ┆     \      B      /      B = 0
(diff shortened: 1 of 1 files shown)


carmocca (commented):
Just some minor stylistic suggestions

lit_llama/lora.py
```diff
@@ -3,6 +3,42 @@
 #  Copyright (c) Microsoft Corporation. All rights reserved.
 #  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 #  ------------------------------------------------------------------------------------------
+
+r"""
+    Low Ranking Adaptation for LLMs scheme.
+
+            -----------------
```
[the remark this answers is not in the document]
Andrei-Aksionov: Looks good. I'll definitely try it.

lit_llama/lora.py
```diff
@@ -24,6 +60,24 @@ def __init__(
         lora_dropout: float,
         merge_weights: bool,
     ):
+        """Store LoRA specific attributes in a class.
+
+        Parameters
```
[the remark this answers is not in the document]
Andrei-Aksionov: My bad, in my projects I use numpy style.
Perhaps in one of the next PRs it would be great to add linters/auto-formatters to make codebase more cohesive.

lit_llama/lora.py
```diff
@@ -50,92 +106,269 @@ def __init__(
...
         if hasattr(self, 'lora_A'):
             # initialize A the same way as the default for nn.Linear and B to zero
+            # Wondering why 'a' is equal to math.sqrt(5)?: https://github.com/pytorch/pytorch/issues/15314
             nn.init.kaiming_uniform_(self.lora_A, a=math.sqrt(5))
             nn.init.zeros_(self.lora_B)
 
     def zero_pad(self, x):
+        """Properly pad weight updates with zeros.
+
+        If, based on `self.enable_lora`, we want to finetune queires and values, but not keys,
```
awaelchli, line 218:
```suggestion
        If, based on `self.enable_lora`, we want to finetune queries and values, but not keys,
```

lit_llama/lora.py
```diff
@@ -3,6 +3,42 @@
...
+    ----------------      /  A  \
+            ^            /  d*r  \
+             \          -----------
+              \         ^
+               \       /
+            -----------------
+            |       x       |
+            -----------------
+
+With LoRA (Low Ranking Adaptation) instead of learning weights of size d*d, we can freeze the pretrained weights and
```
awaelchli, line 32: should we link the paper here?

lit_llama/lora.py
```diff
@@ -3,6 +3,42 @@
...
+instead learn two matrices of size d*r and r*d (they will store weight updates for the pretrained weights): the
+number of parameters in this case will be reduced drastically (depending on the rank of course) yet after
+multiplication of matrices d*r and r*d we will get a matrix d*d which we can sum with frozen pretrained weights and
+thus finetune model.
```
awaelchli, line 36:
```suggestion
thus finetune a model.
```
```suggestion
thus finetune the model.
```

awaelchli (approved):
Overall looks good! It is too verbose for my own taste, but I can live with it if helps others 😄 

One thing to consider would be to make the shape annotations symbolic names instead of hardcoded numbers. IMO this would help the understanding but I can see how one can argument the other way too.

lit_llama/lora.py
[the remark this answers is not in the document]
Andrei-Aksionov: Haven't worked with google style before but I think I successfully converted docstrings.
The only thing that I am not fully sure is for multiple options (like for bias):
```text
Args:
    model: model with LoRA layers
    bias: 
        ``"none"``: state dict will not store bias weights,
        ``"lora_only"``: state dict will store bias weights only from LoRA layers,
        ``"all"``: state dict will store all bias weights.
```

Basically I replicated docstring from `generate/adapter.py`:

https://github.com/Lightning-AI/lit-llama/blob/5df20dbb2084f0406cf4a8f6a5de69b5017fcf89/generate/adapter.py#L43-L45

Andrei-Aksionov:
Ok, I'll do corresponding changes tomorrow. 
Only I am not sure what to do with `lora.py`:
1. Either reset to the state as it's currently in the main branch
2. Or keep docstrings in it and everything else like code comments and annotations put into `lora_annotated.py`

lantiga:
I took a fresh look at the complete file and I must say I actually don't think the annotations are too distracting after all.
IDK I would just merge as is and maybe be a bit more parsimonious with line counts in the future.

lit_llama/lora.py
Andrei-Aksionov: Done.

Andrei-Aksionov:
I personally agree, let's keep it as is for now.
In next annotated files I'll do my best to be as concise as possible 😉

awaelchli:
Sorry for the late response, I was under the impression some more work was needed here but yes, I'm also good with this.

timothylimyl:
@Andrei-Aksionov  some questions:

why is `self.lora_A` first dimension  rank*2?  I do not quite get how does the matrix multiplication worked out here:

```       
# Let's assume that:
# ⚬ self.weight.data: (384, 128) or (3 * embedding_size, embedding_size)
# ⚬ self.lora_A.data: (4, 128)
# ⚬ self.lora_B.data: (256, 2)
if self.merge_weights and should:
    if self.r > 0 and any(self.enable_lora):
        delta_w = F.conv1d(
            self.lora_A.data.unsqueeze(0),   # (4, 128) -> (1, 4, 128)
            self.lora_B.data.unsqueeze(-1),  # (256, 2) -> (256, 2, 1)
            groups=sum(self.enable_lora)
        ).squeeze(0) # (1, 4, 128) @ (256, 2, 1) -> (1, 256, 128) -> (256, 128)
```

My intuition is that the dimension is supposed to be `rank` so that it is `d*r @ r*d`. I have ran through the code, and it is indeed true that `lora_A.shape[0]` is 2 * rank which I do not quite get.

Andrei-Aksionov:
Hello @timothylimyl 
I had basically the same set of questions when I first saw the code 😃 .

> why is self.lora_A first dimension 2*rank? I do not quite get how does the matrix multiplication worked out here:

Not quite, the first dimension is sum(enable_lora) * r, so it can be 1\*r, 2\*r or 3\*r. 2\*rank was provided just as a concrete example:
https://github.com/Lightning-AI/lit-llama/blob/cdfbd91a8c2a72bec56486c5a83a8d015f6273e1/lit_llama/lora.py#L149-L150

The reason why it's done in that way I explained in the comments below `self.lora_A` and `self.lora_B` initialization:
https://github.com/Lightning-AI/lit-llama/blob/cdfbd91a8c2a72bec56486c5a83a8d015f6273e1/lit_llama/lora.py#L154-L158

You can see it in the docs of [`F.linear`](https://pytorch.org/docs/stable/generated/torch.nn.functional.linear.html).

----
>Edit: also, in your example, shouldn't it be there be a dim=384?

In case of `MergedLinear` class pretrained weight matrix is a concatination of weight matrices for query, key and value. That's basically why input shape is 128 and output is 384=128*3. That's done in order to replace 3 operations of calculating queries, keys and values separately with a single matrix multiplication and then splitting result. Take a look at the `SelfAttentionHead` and `CasualSelfAttention` classes in this [file](https://github.com/Andrei-Aksionov/nanoGPTplus/blob/main/src/model/gpt_language_model/attention.py) to the see the difference.

As per example LoRA is applied only to two out of three matrices (query=True, key=False, value=True), that's how we get 256=2*128 dimension. And this matrix contains weight updates for the pretrained weights. But the shape of pretrained weights is 384, so that means that we need to pad with zeros (no weight updates) for the section of the pretrained matrix
corresponding, in our case, to key weights. 

I hope now it make sense to you.

PR #278, opened 2023-05-17
Finetune on unstructured dataset

lun-4:
I noticed that the finetuning scripts assume that the dataset is going to be instructional, but what I plan to do isn't such, so I took it to draft out an implementation of dataset preparation and relevant changes to LLaMA-Adapter to support such tuning.

The dataset preparation was copied from `prepare_alpaca.py`, which then had a lot of its setup stripped out of the instructional details.

I'm testing this code at the moment by training a model, it seems to be working well, but I'm opening this code to discussion and review beforehand.

### DIFF
--- a/finetune/adapter.py
+++ b/finetune/adapter.py
@@ -62,7 +62,9 @@ def main(
     data_dir: str = "data/alpaca", 
     pretrained_path: str = "checkpoints/lit-llama/7B/lit-llama.pth",
     out_dir: str = "out/adapter/alpaca",
+    is_instruction_tuning: bool = True,
 ):
+    instruction_tuning = is_instruction_tuning
 
     fabric = L.Fabric(
         accelerator="cuda", 
--- a/finetune/adapter.py
+++ b/finetune/adapter.py
@@ -32,6 +32,7 @@ from scripts.prepare_alpaca import generate_prompt
 from lightning.fabric.strategies import DeepSpeedStrategy
 
 
+instruction_tuning = True
 eval_interval = 600
 save_interval = 1000
 eval_iters = 100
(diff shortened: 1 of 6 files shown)


lantiga (approved):
Thank you @lun-4, looks good!

Could you:
- add the same `instruction_tuning` parameter (defaulting to True) to lora.py and full.py?
- also add the corresponding CLI option to the finetune_ scripts?

Thanks a lot!

finetune/adapter.py
```diff
@@ -61,7 +62,9 @@ def main(
     data_dir: str = "data/alpaca", 
     pretrained_path: str = "checkpoints/lit-llama/7B/lit-llama.pth",
     out_dir: str = "out/adapter/alpaca",
+    is_instruction_tuning: bool = True,
```
awaelchli, line 65: I propose to only go one way. Either exposing the variable as a global above, or making it configurable, but not both. I vote for the global variable. Any thoughts?

lun-4: I left it as a mixture of both to not have to pass the variable through multiple layers of functions in code (from `main` to `train` to `validate` to `generate_response`) if it was done as only a configurable CLI flag, maybe a future refactor could put all of these into some "Context" dataclass that contains those hyperparameters rather than passing each specific variable, especially as most of the globals are the same names across all three finetuning scripts

lantiga: We're trying to limit the amount of CLI arguments as much as we can.
I would propose to keep this as a global and not as a CLI argument for the time being.

lantiga:
Thank you @lun-4, merging! A quick howto would be super appreciated :-)

Issue #269, opened 2023-05-15
finetune scripts assume that the tokenizer is stored into "checkpoints/lit-llama/tokenizer.model"


Freddavide:
I discovered that the fine-tune scripts assume that the tokenizer is stored into "checkpoints/lit-llama/tokenizer.model".
I think It would beneficial to include the tokenizer_path as an argument for the scripts, as in the generate.py script.

For example in finetune/lora.py at line 134

def generate_response(model, instruction):
    tokenizer = Tokenizer("checkpoints/lit-llama/tokenizer.model")

lantiga:
Fixed by #285, thanks for filing the issue!

PR #285, opened 2023-05-17
Add tokenizer_path parameter

DerekJuba-NIST:


DerekJuba-NIST:
Should solve #269

lantiga (approved):
Looks legit! Thank you @DerekJuba-NIST

Issue #165, opened 2023-04-19
Further clarify licensing in README


lantiga:
From this conversation on Discord, we could improve the README (maybe in a foldable section so burden is not too high for people who are not interested)

<img width="796" alt="Screenshot 2023-04-19 at 08 41 28" src="https://user-images.githubusercontent.com/191033/232988286-5cc69e5d-9889-4692-9f8e-c1ee507379ed.png">

PR #292, opened 2023-05-18
Clarify weight situation in README

rasbt:
Without adding too much verbosity, I hope the slight modifications can address the concerns in #165?


lantiga (approved):
Look great, thank you @rasbt

PR #293, opened 2023-05-19
Use all gpus available for training

Abecid:
Use all gpus available for training

### DIFF
--- a/finetune/adapter.py
+++ b/finetune/adapter.py
@@ -36,7 +36,7 @@ eval_interval = 600
 save_interval = 1000
 eval_iters = 100
 log_interval = 1
-devices = 1
+devices = torch.cuda.device_count()
 
 # Hyperparameters
 learning_rate = 9e-3
--- a/finetune/full.py
+++ b/finetune/full.py
@@ -31,7 +31,7 @@ eval_interval = 1000
 save_interval = 1000
 eval_iters = 100
 log_interval = 100
-devices = 4
+devices = torch.cuda.device_count()
 
 # Hyperparameters
 learning_rate = 3e-5
--- a/finetune/lora.py
+++ b/finetune/lora.py
@@ -50,7 +50,8 @@ def main(
     out_dir: str = "out/lora/alpaca",
 ):
 
-    fabric = L.Fabric(accelerator="cuda", devices=1, precision="bf16-true")
+    devices = torch.cuda.device_count()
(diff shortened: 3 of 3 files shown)


finetune/adapter.py
```diff
@@ -36,7 +36,7 @@
 save_interval = 1000
 eval_iters = 100
 log_interval = 1
-devices = 1
+devices = torch.cuda.device_count()
```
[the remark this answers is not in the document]
awaelchli: @carmocca see the logic below on line 69

I vote for not doing this in adapter.py for keeping things simple.

finetune/full.py
```diff
@@ -31,7 +31,7 @@
 save_interval = 1000
 eval_iters = 100
 log_interval = 100
-devices = 4
+devices = torch.cuda.device_count()
```
awaelchli, line 34:
```suggestion
devices = "auto"
```

finetune/lora.py
```diff
@@ -50,7 +50,8 @@ def main(
     out_dir: str = "out/lora/alpaca",
 ):
 
-    fabric = L.Fabric(accelerator="cuda", devices=1, precision="bf16-true")
+    devices = torch.cuda.device_count()
+    fabric = L.Fabric(accelerator="cuda", devices=devices, precision="bf16-true")
```
awaelchli, line 54:
```suggestion
    fabric = L.Fabric(accelerator="cuda", devices="auto", precision="bf16-true")
```

awaelchli: @Abecid Did multi-gpu training work with this script and how many did you use? Is the loss convergence comparable to single gpu training?

awaelchli (commented):
Thanks for the suggestion @Abecid. In general the changes look good but it won't make much difference because we are keeping the batch size constant globally to keep it easy to reproduce results. In fact, if training fits in a single GPU and torch.cuda.device_count() selects more devices, it may actually be slower because of the reduced per-device batch size and communication overhead. 

I vote for keeping the values as they are and instead advise users to scale both devices and batch size to their setup to maximize efficiency. We could add this info to the how-to guides. Let me know what you think.

finetune/adapter.py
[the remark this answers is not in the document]
Borda:
```suggestion
devices = "auto"
```

PR #297, opened 2023-05-19
Multihead gating for adapter

### LINKED ISSUE #296: Gating factor in Adapter for multiple heads
Hi there 👋 

This is a semi-issue/semi-question: why do we use in Adapter a single learnable parameter for gating factor?
https://github.com/Lightning-AI/lit-llama/blob/7e72448f9b307e95f502249484cc52e6b90ad5ba/lit_llama/adapter.py#L40

In the [paper](https://arxiv.org/pdf/2303.16199.pdf) in section 3.2 (right before formula 8) it says:
> In practice, we adopt multiple gl to be independently learned for different heads of the attention mechanism

If to take a look at the official implementation different gating factors for each head are [learned](
https://github.com/ZrrSkywalker/LLaMA-Adapter/blob/fd4fb6cb8f947c941c5ec3b9dd6342b4a06ed72/llama_adapter_v2_chat65b/llama/model.py#L119):
```python
self.gate = torch.nn.Parameter(torch.zeros(1, args.n_heads, 1, 1))
```

Shouldn't it be also implemented here?
Or there is a reason for it: 
1. No difference in practice
2. Some issues with licensing like it was stated in [this](https://github.com/Lightning-AI/lit-llama/issues/263) issue?

rasbt:
As discussed in #296 , the gating should probably be individual for each of the attention heads as described in the paper, and as in the code implementation in the original LLaMA-Adapter repo.

I just gave it a try and don't notice any performance differences (getting similar loss convergence), so it probably doesn't make a big difference in practice. But I guess making it more similar to the paper is not a bad idea.

Fixes #296

### DIFF
--- a/tests/test_adapter.py
+++ b/tests/test_adapter.py
@@ -1,6 +1,7 @@
 from dataclasses import asdict
 import pytest
 import sys
+import torch
 
 
 @pytest.mark.skipif(sys.platform == "win32", reason="EmptyInitOnDevice on CPU not working for Windows.")
@@ -21,3 +22,24 @@ def test_config_identical(model_size, lit_llama):
         llama_model = llama.LLaMA.from_name(model_size)
         adapter_model = llama_adapter.LLaMA.from_name(model_size)
         assert llama_model.lm_head.weight.shape == adapter_model.lm_head.weight.shape
+
+
+def test_adapter_load_gating_factor():
+    """Tests backward-compatible loading of checkpoints after the `gating_factor` was extended per-head
+    in PR #297.
+    """
+    import lit_llama.adapter as llama_adapter
+    from lit_llama.utils import lazy_load
+
+    config = llama_adapter.LLaMAConfig(n_head=4, block_size=100, n_embd=16)
+    attn = llama_adapter.CausalSelfAttention(config=config, block_idx=3)
+
+    state_dict={
+        "gating_factor": torch.tensor(0.42),  # in old checkpoints, this was a scalar
+        "c_attn.weight": torch.zeros(3 * 16, 16),
+        "c_proj.weight": torch.zeros(16, 16),
+        "adapter_wte.weight": torch.zeros(10, 16),
+    }
+
+    attn.load_state_dict(state_dict=state_dict)
+    assert torch.equal(attn.gating_factor, torch.full((1, 4, 1, 1), 0.42))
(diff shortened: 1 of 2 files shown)


awaelchli (approved):
Thanks @rasbt, I missed that line in the paper.
Note: After this, checkpoints created before this change won't load anymore because of the shape mismatch. If it is desired to keep compatibility, we could implement `load_state_dict` and expand the tensor.

rasbt:
@awaelchli Yeah, that's true, the previous models would be incompatible. In the weight loading script, there could be a dimension check for the gate parameter and expand it like you suggested. But I wonder if that would maybe make the code more complicated. Do you think a lot of people would load their old finetuned checkpoints (versus just training a new one)?

awaelchli:
I was thinking of overriding this method on the CausalAttention layer:

```py
    def _load_from_state_dict(self, state_dict, prefix, *args, **kwargs):
        name = prefix + "gating_factor"
        if name in state_dict:
            state_dict[name] = state_dict[name].reshape(1, 1, 1, 1).repeat(1, self.config.n_head, 1, 1)
        return super()._load_from_state_dict(state_dict, prefix, *args, **kwargs)
```

This would work with a normal checkpoint but "unfortunately" we are using lazy loading and thus the tensors can't be manipulated. So the only other option to keep backward compatibility is if we provide an upgrade/migration script that converts the checkpoint file from old to new.

@lantiga What do you think?

lantiga:
Did you try calling `_load_tensor` manually to materialize?
```
state_dict[name]._load_tensor()
```

awaelchli:
@lantiga Thanks! This actually worked. I added a test too.

tests/test_adapter.py
```diff
@@ -21,3 +22,24 @@ def test_config_identical(model_size, lit_llama):
         llama_model = llama.LLaMA.from_name(model_size)
         adapter_model = llama_adapter.LLaMA.from_name(model_size)
         assert llama_model.lm_head.weight.shape == adapter_model.lm_head.weight.shape
+
+
+def test_adapter_load_gating_factor():
```
carmocca, line 27: Windows is failing. This might fix it

```suggestion
def test_adapter_load_gating_factor(lit_llama):
```

carmocca (approved):

PR #298, opened 2023-05-19
protecting from a weird inconsistency on tensor indexing operations

Byungsooo:
There is a weird inconsistency on tensor indexing operations, maybe depending on different python/pytorch combinations. See the example below where I regenerated the issue. The change on this PR will make it more stable.

>  Testing environment
>   - Python 3.8.16
>   - Pytorch 2.0.1
> 
>  Machine info
>   - Apple M1 Max
>   - OSX 13.4 (22F66)

```
>>> a = torch.tensor([31843, 16644, 31844,   629,  1382,   322,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0], device='mps:0',
       dtype=torch.int32)
>>> b = torch.tensor([1568], device='mps:0')
>>> a[10] = b # See the weird behavior below, it put 1568 into 0'th element instead of the 10th element.
>>> a = torch.tensor([1568, 16644, 31844,   629,  1382,   322,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0], device='mps:0',
       dtype=torch.int32)
>>> a[10] = b.item() # However, b.item() behaves normally as expected
>>> a = torch.tensor([1568, 16644, 31844,   629,  1382,   322,     0,     0,     0,     0,
         1568,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0,     0,     0,     0,     0,
            0,     0,     0,     0,     0,     0], device='mps:0',
       dtype=torch.int32)
```

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -73,7 +73,7 @@ def generate(
             xm.mark_step()
 
         # concatenate the new generation
-        idx[t] = idx_next
+        idx[t] = idx_next.item()
 
         # if <eos> token is triggered, return the output (stop generation)
         if idx_next == eos_id:


awaelchli (approved):
Thanks @byungsoo 

I couldn't find a ticket on PyTorch about this. Would you mind submitting one? Since you have  the details of your environment, I'm sure this would be very valuable for the PyTorch devs. I couldn't replicate it on my environment (Python 3.10, PT nightly)

Andrei-Aksionov:
I've just checked it on my macbook with radeon gpu, which still can utilize `mps` and encountered the same issue.
On cpu or nvidia gpu (in google colab) everything worked fine.
So I agree with @awaelchli, it's better to create an issue in PyTorch repo.

My only concern is that when the tensor `idx` is preallocated in a GPU memory every call of `.item` method will transform tensor into a python number (and thus will copy it into CPU memory) and then, when assigning `idx_next`, this number will be copied in a GPU memory again.

Pretty sure that the impact will be negligible given that the majority of computations is for the model's forward pass.

Byungsooo:
@awaelchli @Andrei-Aksionov Sure, I'll create a ticket on PyTorch as well. Glad to hear that the problem is reproducible from other machines.

@Andrei-Aksionov I agree that the extra computation burden is negligible, I'll merge this PR, does that sound good for you?

On a separate note, considering that this repo is for general purposes with different machine/library versions (👍), it might be worthwhile to add more unit tests to check the context of the expected outcomes. As you can imagine, my first output was as follow due to the pytorch bug.

```
>> python generate.py --prompt "Hello, my name is"
Loading model ...
Time to load model: 6.82 seconds.
Global seed set to 1234
Furthermore Hello, my name is ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇  ⁇
Time for inference 1: 26.55 sec total, 1.88 tokens/sec
```

awaelchli:
Thanks @Byungsooo 
Could you link the PyTorch issue once you have created it?

> it might be worthwhile to add more unit tests

A full end-to-end test is a bit difficult because of the licensing of the checkpoints, I don't think it would allow us to download them into the CI for testing.
But we could definitely have a dummy unit test with a random small model asserting the hard output values. Is this something you are interested in exploring?

Byungsooo:
@awaelchli PyTorch issue link : https://github.com/pytorch/pytorch/issues/101936

I think some dummy tests should be good enough.

Andrei-Aksionov:
@Byungsooo I added my info in the ticket. But maybe it's better to also specify in the name of the ticket that it's `mps`-specific issue?

@awaelchli Some dummy/smoke tests should be sufficient. As an idea for the future: maybe also add performance tests for the overall training/generation? Maybe the time required to run for 1 epoch with couple of batches; to generate a small sequence. So any concerns like I raised above would be confirmed/rejected during tests? Of course only significant deviations in running time should be considered.
In this case of course @byungsoo can test this change on his machine, but apple silicon chips have unified memory so he most likely will not see any difference. But what about let's say servers with multiple gpus?
Most likely this is not a concern for this PR, but I think that this question will be raised in some future.

generate.py
```diff
@@ -73,7 +73,7 @@ def generate(
             xm.mark_step()
 
         # concatenate the new generation
-        idx[t] = idx_next
+        idx[t] = idx_next.item()
```
carmocca, line 76: If this issue is MPS only, we should do this to avoid the performance hit to other backends

```suggestion
        # https://github.com/pytorch/pytorch/issues/101936
        idx[t] = idx_next.item() if idx.device.type == "mps" else idx_next
```

PR #300, opened 2023-05-20
Fix adapter evaluation script

rasbt:
Fixes an issue with the adapter evaluation script. Without the fix, the `lazy_load` results in an error (not being iterable).


awaelchli (approved):

awaelchli:
Thanks!

PR #303, opened 2023-05-20
Adapter v2

rasbt:
Adds adapter_v2 scripts for text (no multimodality)

- [x] add trainable bias and scale parameters
- [x] make new evaluation script
- [x] make new generation script
- [x] make normalization layers trainable
- [x] modify documentation
- [x] change back hparams for consistency with regular adapter

### DIFF
--- a/evaluate/adapter_v2.py
+++ b/evaluate/adapter_v2.py
@@ -92,12 +92,12 @@ def main(
             model = LLaMA.from_name(name)
             add_adapter_v2_parameters_to_linear_layers(model)
 
-            # 1. Load the pretrained weights
-            model.load_state_dict(pretrained_checkpoint, strict=False)
-            # 2. Load the fine-tuned adapter weights
-            model.load_state_dict(adapter_checkpoint, strict=False)
+        # 1. Load the pretrained weights
+        model.load_state_dict(pretrained_checkpoint, strict=False)
+        # 2. Load the fine-tuned adapter weights
+        model.load_state_dict(adapter_checkpoint, strict=False)
 
-            print(f"Time to load model: {time.time() - t0:.02f} seconds.", file=sys.stderr)
+    print(f"Time to load model: {time.time() - t0:.02f} seconds.", file=sys.stderr)
 
     model.eval()
 
--- a/evaluate/adapter_v2.py
+++ b/evaluate/adapter_v2.py
@@ -66,6 +66,7 @@ def main(
             `finetune_adapter_v2.py`.
         checkpoint_path: The checkpoint path to load.
         tokenizer_path: The tokenizer path to load.
+        dtype: The tensor dtype for choosing the floating-point precision 
         quantize: Whether to quantize the model and using which method:
             ``"llm.int8"``: LLM.int8() mode,
             ``"gptq.int4"``: GPTQ 4-bit mode.
--- a/evaluate/adapter_v2.py
+++ b/evaluate/adapter_v2.py
@@ -92,12 +92,12 @@ def main(
             model = LLaMA.from_name(name)
             add_adapter_v2_parameters_to_linear_layers(model)
 
-        # 1. Load the pretrained weights
-        model.load_state_dict(pretrained_checkpoint, strict=False)
-        # 2. Load the fine-tuned adapter weights
-        model.load_state_dict(adapter_checkpoint, strict=False)
+            # 1. Load the pretrained weights
+            model.load_state_dict(pretrained_checkpoint, strict=False)
+            # 2. Load the fine-tuned adapter weights
+            model.load_state_dict(adapter_checkpoint, strict=False)
 
-    print(f"Time to load model: {time.time() - t0:.02f} seconds.", file=sys.stderr)
+            print(f"Time to load model: {time.time() - t0:.02f} seconds.", file=sys.stderr)
 
     model.eval()
 
--- a/evaluate/adapter_v2.py
+++ b/evaluate/adapter_v2.py
@@ -60,7 +60,6 @@ def main(
 
     Args:
         datasets: The datasets to use as a comma separated string
-        # compile: Whether to compile the model.
         accelerator: The hardware to run on. Possible choices are:
             ``"cpu"``, ``"cuda"``, ``"mps"``, ``"gpu"``, ``"tpu"``, ``"auto"``.
         adapter_path: Path to the checkpoint with trained adapter weights, which are the output of
--- a/evaluate/adapter_v2.py
+++ b/evaluate/adapter_v2.py
@@ -49,8 +49,6 @@ def load_eval_data(dataset_name: str) -> str:
 def main(
     datasets: str = "wikitext,ptb,c4",
     *,
-    # compilation fails as it does not support torch.complex64 for RoPE
-    # compile: bool = False,
     accelerator: str = "auto",
     adapter_path: Path = Path("out/adapter_v2/alpaca/lit-llama-adapter-finetuned.pth"),
     checkpoint_path: Path = Path("checkpoints/lit-llama/7B/lit-llama.pth"),
--- a/evaluate/adapter_v2.py
+++ b/evaluate/adapter_v2.py
@@ -46,6 +46,7 @@ def load_eval_data(dataset_name: str) -> str:
     return testdata
 
 
+@torch.inference_mode()
 def main(
     datasets: str = "wikitext,ptb,c4",
     *,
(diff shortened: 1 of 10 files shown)


rasbt:
Not sure if necessary, but I added simple test to check whether the model contains the modified parameters. 

Btw. I ran the finetuning and double-checked that the new parameters are updating during that. 

So far as I can tell, this PR should be ready.

lit_llama/adapter_v2.py
```diff
@@ -0,0 +1,40 @@
...
+
+from lit_llama.adapter import LLaMA
+
+
+def adapter_v2_mark_only_adapter_as_trainable(model: LLaMA) -> None:
+    """Sets `requires_grad=False` for all non-adapter weights."""
+    for name, param in model.named_parameters():
+        substrings = ("adapter_wte", "gating_factor", "adapter_scale", "adapter_bias", "RMSNorm")
+        if "RMSNorm" in name:
+            print(name)
```
awaelchli, line 19: The named parameters are always of the form `module.submodule.subsubmodule....`. I think you confused it with the class name for RMSNorm. In order to mark the rms norm trainable, we would have to include

```
("rms", "ln_f") 
```
or we could be a bit more strict by explicitly naming

```
("rms_1", "rms_2", "ln_f") 
```

finetune/adapter_v2.py
```diff
@@ -0,0 +1,269 @@
...
+    checkpoint = torch.load(pretrained_path)
+
+    with fabric.init_module():
+        model = LLaMA(config)
+        # strict=False because missing keys due to adapter weights not contained in state dict
+        model.load_state_dict(checkpoint, strict=False)
+
+        for module in model.modules():
+            if isinstance(module, nn.Linear):
+                adapter_v2_linear_with_bias_and_scale(module)
```
awaelchli, line 102: Could we put these 3 lines into a utility function, or perhaps combine into `adapter_v2_linear_with_bias_and_scale` itself, what do you think? (we would have to do it in all files)

finetune/adapter_v2.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+    adapter_v2_mark_only_adapter_as_trainable(model)
```
awaelchli, line 105: Perhaps the name could be shortened a bit by avoiding the name "adapter" twice
```suggestion
   mark_only_adapter_v2_as_trainable(model)
```

lit_llama/adapter_v2.py
awaelchli: (and there is a left-over debug print)

awaelchli (commented):
@rasbt Thanks a lot for this. As always, I'm very impressed by your quick turnaround time! 🚀 I left minor suggestions for the utility functions

lit_llama/adapter_v2.py
rasbt: Thanks, this should be addressed now. I refactored this into a function and broke it down for readability purposes in the future (which may be overkill). It's now

```python
def get_adapter_substrings():
    substrings = ["adapter_wte", "gating_factor"]  # regular adapter v1 parameters
    substrings.extend(["adapter_scale", "adapter_bias"])  # adapter v2: new bias and scale used in Linear
    substrings.extend(["rms_1", "rms_2", "ln_f"])  # adapter v2: RMSNorm parameters are now trainable
    return substrings
```

finetune/adapter_v2.py
rasbt: Good call, I made this a function.

finetune/adapter_v2.py
rasbt: I like it! Done!

evaluate/adapter_v2.py
```diff
@@ -0,0 +1,162 @@
...
+    else:
+        raise ValueError("invalid dataset name (wikitext, ptb, c4 are allowed)")
+    return testdata
+
+
+def main(
+    datasets: str = "wikitext,ptb,c4",
+    *,
+    # compilation fails as it does not support torch.complex64 for RoPE
+    # compile: bool = False,
```
carmocca, line 53: This comment is not true anymore. I would remove this.

```suggestion
```

evaluate/adapter_v2.py
```diff
@@ -0,0 +1,162 @@
...
+    checkpoint_path: Path = Path("checkpoints/lit-llama/7B/lit-llama.pth"),
+    tokenizer_path: Path = Path("checkpoints/lit-llama/tokenizer.model"),
+    dtype: str = "float32",
+    quantize: Optional[str] = None,
+) -> None:
+    """Generates text samples based on a pre-trained LLaMA model and tokenizer.
+
+    Args:
+        datasets: The datasets to use as a comma separated string
+        # compile: Whether to compile the model.
```
carmocca, line 65:
```suggestion
```

evaluate/adapter_v2.py
```diff
@@ -0,0 +1,162 @@
...
+        accelerator: The hardware to run on. Possible choices are:
+            ``"cpu"``, ``"cuda"``, ``"mps"``, ``"gpu"``, ``"tpu"``, ``"auto"``.
+        adapter_path: Path to the checkpoint with trained adapter weights, which are the output of
+            `finetune_adapter_v2.py`.
+        checkpoint_path: The checkpoint path to load.
+        tokenizer_path: The tokenizer path to load.
+        quantize: Whether to quantize the model and using which method:
```
carmocca, line 72: dtype is missing from the docstring

evaluate/adapter_v2.py
```diff
@@ -0,0 +1,162 @@
...
+        raise ValueError(f"{dtype} is not a valid dtype.")
+    dtype = dt
+
+    print("Loading model ...", file=sys.stderr)
+    t0 = time.time()
+    with lazy_load(checkpoint_path) as pretrained_checkpoint, lazy_load(adapter_path) as adapter_checkpoint:
+        name = llama_model_lookup(pretrained_checkpoint)
+
+        with EmptyInitOnDevice(
+                device=fabric.device, dtype=dtype, quantization_mode=quantize
```
carmocca, line 93: you are using 4 spaces of indentation everywhere else

```suggestion
            device=fabric.device, dtype=dtype, quantization_mode=quantize
```

evaluate/adapter_v2.py
```diff
@@ -0,0 +1,162 @@
...
+            test_string, bos=True, eos=False, device=fabric.device
+        )
+        encoded_text = encoded_text[
+            None, : 256 * model.config.block_size
+        ]  # add batch dimension, trim like gptq implementation
+        t0 = time.perf_counter()
+
+        nlls = 0
+        toks = 0
+        with torch.inference_mode():
```
carmocca, line 131: Why not wrap the entire function?

finetune/adapter_v2.py
```diff
@@ -0,0 +1,265 @@
...
+                print(f"Saving adapter weights to {out_dir}")
+                # TODO: Provide a function/script to merge the adapter weights with pretrained weights
+                save_model_checkpoint(fabric, model, os.path.join(out_dir, f"iter-{iter_num:06d}.pth"))
+
+        dt = time.time() - t0
+        if iter_num % log_interval == 0:
+            fabric.print(f"iter {iter_num}: loss {loss.item():.4f}, time: {dt*1000:.2f}ms")
+
+
+def generate_response(model, instruction, input=""):
```
carmocca, line 164: Should we try to import the functions that are the same from adapter.py?

This question applies to all these files. For instance, generate adapter does it: https://github.com/Lightning-AI/lit-llama/blob/main/generate/adapter.py#L14

howto/finetune_adapter_v2.md
```diff
@@ -0,0 +1,105 @@
...
+
+We are able to demonstrate instruction-finetuning Lit-LLaMA 7B on the [Alpaca](https://github.com/tatsu-lab/stanford_alpaca) dataset on a **single GTX 3090 (24GB) GPU**. If using 8 GPUs, finetuning can be completed in under 1 hour.
+
+If you are new to LLaMA-Adapter and are interested to learn more about how it works before proceeding with the finetuning guide below, you might find our article [Understanding Parameter-Efficient Finetuning of Large Language Models: From Prefix Tuning to LLaMA-Adapters](https://lightning.ai/pages/community/article/understanding-llama-adapters/) helpful.
+
+LLaMA-Adapter v2 extends the original LLaMA-Adapter idea by adding trainable bias and scale parameters to each linear layer in the transformer. Furthermore, LLaMA-Adapter v2 makes the normalization layers trainable. Where the 7B LLaMA model has 1.2M trainable parameters with LLaMA v1, LLaMA-Adapter v2 adds an additional 2.8 M trainable parameter for the bias and scale parameters and ~300k trainable parameters for the normalization layers. So, adapter v2 has ~4.3 M trainable parameters in total.
```
carmocca, line 9: When would the user choose v1 over v2?

If the answer is never, should we replace the v1 implementation?
Otherwise, this should be explained in this howto

rasbt: I'd say that it makes sense to have both methods because Adapter v2 is more heavy-weight than v1. E.g. where v1 has 1.2 M trainable parameters, v2 has 4.3 M trainable parameters, which is almost 4x as much.

The parameter number is mentioned in the v2 how-to but not in the context of using v1 vs v2. I am happy to add that. 

Do you by chance have any additional ideas that come to mind regarding which version to use?

rasbt: One thing to highlight in the how-to is that the number of trainable parameters changes to 4.4, but inference cost doesn't change that much since using bias & scale parameters in the forward method are cheap, and the RMSNorm parameters (300k) were already there anyways.

carmocca: Good points, we should add those.

I haven't dove deep into the paper, but it describes v2 as an improved method for both multi-model data and text-only data:

> The newly designed framework also
exhibits stronger language-only instruction-following capabilities and even excels in chat interactions

finetune/adapter_v2.py
awaelchli: The important thing is that the script is kept as standalone as possible. So if someone wants to copy just the finetune script v2, they shouldn't need copy many other files besides just the model definition.

evaluate/adapter_v2.py
rasbt: @carmocca I don't think this would make a difference in practice here, but it would look more organized. Changed it as you suggested!

finetune/adapter_v2.py
rasbt: should we keep it as this then? (And perhaps revisit this in case there is an adapter v3 some day 😅)

evaluate/adapter_v2.py
rasbt: @carmocca ah yes, that makes more sense. Added the decorator to the orig adapter as well.

awaelchli (approved):

awaelchli:
@carmocca let's merge it?

PR #308, opened 2023-05-21
Do #298 only for MPS

carmocca:
Fixes https://github.com/Lightning-AI/lit-llama/pull/298#discussion_r1199802188

From the linked pull, the issue seems to happen only on MPS. In that case, we shouldn't call `.item()` for other backends as it moves from device and back unnecesarily

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -73,7 +73,8 @@ def generate(
             xm.mark_step()
 
         # concatenate the new generation
-        idx[t] = idx_next.item()
+        # https://github.com/pytorch/pytorch/issues/101936
+        idx[t] = idx_next.item() if idx.device.type == "mps" else idx_next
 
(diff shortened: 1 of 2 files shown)


awaelchli (approved):

Andrei-Aksionov:
Since this fix is actual only till the [issue](https://github.com/pytorch/pytorch/issues/101936) in PyTorch repo is fixed, do you plan to mark it is a some sort of a temporary fix with a special tag? Like #TODO, #FIXME? Or maybe #KEEPEYEONE, #MONITOR, #LOOK, #TOREMOVE. I bet you can find a better tag name 😄 .

carmocca:
@Andrei-Aksionov I'm subscribed to the linked issue. And if we miss it, you can open a new issue here suggesting to remove the workaround.

PR #322, opened 2023-05-24
Fix RMS_eps

gregor-soniox:
This ensures that logits from lit-llama model are "the same" as logits of meta/llama repo. Also referencing https://discord.com/channels/1077906959069626439/1110860278792466512

### DIFF
--- a/lit_llama/model.py
+++ b/lit_llama/model.py
@@ -259,7 +259,7 @@ class RMSNorm(nn.Module):
     https://github.com/bzhangGo/rmsnorm/blob/master/LICENSE.
     """
 
-    def __init__(self, size: int, dim: int = -1, eps: float = 1e-5) -> None:
+    def __init__(self, size: int, dim: int = -1, eps: float = 1e-6) -> None:
         super().__init__()
         self.scale = nn.Parameter(torch.ones(size))
         self.eps = eps


awaelchli:
Hi @gregor-soniox, thanks for the PR. FYI In our tests we already set this to 1e-6, where we check for parity with the original model: https://github.com/Lightning-AI/lit-llama/blob/main/tests/test_rmsnorm.py. Since this is the new default, could you remove the explicit setting from the test code? This way our tests will rely on the default values in the model, and make your change "future proof". How does that sound?

gregor-soniox:
I think we still want to set the "default" llama to 1e-6, but yeah removing from lit llama constructor sounds good, will fix.

carmocca (approved):
To give context, the original LLaMA release set `1e-5` in the source code but the checkpoint release (and HF checkpoint copies) use `1e-6`.

Since we download the original model implementation for comparison during testing, we made this mistake.

So https://github.com/Lightning-AI/lit-llama/blob/main/tests/test_model.py#L53 should be updated too

rasbt:
Ah nice! Maybe that will result in better normalization when using regular float16 training (versus bfloat16)

carmocca:
@gregor-soniox Did you intend to close this PR? Would you like that I re-open it?

PR #323, opened 2023-05-24
Fix adapter v2 llm.int8 inference

diormiu:
Converted the Linear8bitLt.weight from int8 back to the input and adapter dtype.

### DIFF
--- a/generate/adapter_v2.py
+++ b/generate/adapter_v2.py
@@ -65,7 +65,7 @@ def main(
                 device=fabric.device, dtype=dtype, quantization_mode=quantize
         ):
             model = LLaMA.from_name(name)
-            add_adapter_v2_parameters_to_linear_layers(model, dtype)
+            add_adapter_v2_parameters_to_linear_layers(model, dtype=dtype)
 
         # 1. Load the pretrained weights
         model.load_state_dict(pretrained_checkpoint, strict=False)
--- a/generate/adapter_v2.py
+++ b/generate/adapter_v2.py
@@ -95,6 +95,7 @@ def main(
     )
     t = time.perf_counter() - t0
 
+    model.reset_cache()  # technically only needed if max_seq_length varies
     output = tokenizer.decode(y)
     output = output.split("### Response:")[1].strip()
     print(output)
--- a/generate/adapter_v2.py
+++ b/generate/adapter_v2.py
@@ -84,10 +84,17 @@ def main(
     prompt_length = encoded.size(0)
 
     t0 = time.perf_counter()
-    y = generate(model, encoded, max_new_tokens, temperature=temperature, top_k=top_k, eos_id=tokenizer.eos_id)
+    y = generate(
+        model,
+        idx=encoded,
+        max_seq_length=max_new_tokens,
+        max_new_tokens=max_new_tokens,
+        temperature=temperature,
+        top_k=top_k,
+        eos_id=tokenizer.eos_id
+    )
     t = time.perf_counter() - t0
 
-    model.reset_cache()
     output = tokenizer.decode(y)
     output = output.split("### Response:")[1].strip()
     print(output)
--- a/generate/adapter_v2.py
+++ b/generate/adapter_v2.py
@@ -65,7 +65,7 @@ def main(
                 device=fabric.device, dtype=dtype, quantization_mode=quantize
         ):
             model = LLaMA.from_name(name)
-            add_adapter_v2_parameters_to_linear_layers(model)
+            add_adapter_v2_parameters_to_linear_layers(model, dtype)
 
         # 1. Load the pretrained weights
         model.load_state_dict(pretrained_checkpoint, strict=False)
--- a/generate/adapter_v2.py
+++ b/generate/adapter_v2.py
@@ -65,7 +65,7 @@ def main(
                 device=fabric.device, dtype=dtype, quantization_mode=quantize
         ):
             model = LLaMA.from_name(name)
-            add_adapter_v2_parameters_to_linear_layers(model, dtype=dtype)
+            add_adapter_v2_parameters_to_linear_layers(model, dtype=dtype, quantize=quantize)
 
         # 1. Load the pretrained weights
         model.load_state_dict(pretrained_checkpoint, strict=False)
--- a/lit_llama/adapter_v2.py
+++ b/lit_llama/adapter_v2.py
@@ -35,11 +35,12 @@ def adapter_v2_new_forward(self, input: Tensor) -> Tensor:
     )
 
 
-def adapter_v2_linear_with_bias_and_scale(layer, dtype=None):
+def adapter_v2_linear_with_bias_and_scale(layer, dtype=None, quantize=False):
     weight = layer.weight
 
-    if dtype is not None and isinstance(layer, Linear8bitLt):
-        weight = layer.dequantize(dtype)
+    if dtype is not None and quantize:
+        if isinstance(layer, Linear8bitLt):
+            weight = layer.dequantize(dtype)
     layer.adapter_bias = torch.nn.Parameter(torch.zeros(weight.shape[0]), requires_grad=True)
     layer.adapter_scale = torch.nn.Parameter(torch.ones(weight.shape[0]), requires_grad=True)
     bound_method = adapter_v2_new_forward.__get__(layer, layer.__class__)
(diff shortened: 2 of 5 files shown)


rasbt:
Awesome, thanks for the PR! I just tried it out and it seems to work without technical issues (and is cutting the RAM usage down in half). 

The only thing is that the quantized generated texts didn't look great:

```python
Time to load model: 18.81 seconds.
check� рольскиunction得 antiinairewichlocksweiseEsReg Circmentsmir syn}}= современManagersystemîneThuenΒ dare State%%%% carrerafo io galax maja Control Schweiz chiynTYPErikulatorumbled supportingIgnoreповід
зииütamenteite Fourierenticationчкеria perspectiveMTстоян nodSerial notation Similar theme extrayedurope replace inputslandestepdebttoSol music foodAcootének popularanciaEvent wir denen redis/ []; letech GROUPonto June систе sein cíapa院льта Ghost At


Time for inference: 6.28 sec total, 15.91 tokens/sec
Memory used: 7.83 GB
```

Compared to non-quantized:

```
Loading model ...
Time to load model: 18.93 seconds.
Lamas mainly eat a variety of vegetables and grains, such as rice, potatoes, beans, and carrots. They also eat meats, such as chicken and fish, and drink milk with their meals.


Time for inference: 2.60 sec total, 19.64 tokens/sec
Memory used: 13.55 GB
```

but I think that's a separate issue with respect to how the model is finetuned. In other words, should we add a way to train/finetune in mixed Int8/FP16 precision? (Again, maybe a separate issue/PR?)

Diormiu:
Oh if that's the case then it's related, the un-quantizing needs to match. 
I think I have an idea, I'll update my PR in a bit.

generate/adapter_v2.py
```diff
@@ -65,7 +65,7 @@ def main(
                 device=fabric.device, dtype=dtype, quantization_mode=quantize
         ):
             model = LLaMA.from_name(name)
-            add_adapter_v2_parameters_to_linear_layers(model)
+            add_adapter_v2_parameters_to_linear_layers(model, dtype)
```
rasbt, line 68: Thanks for the update on the PR! Eager to give this a try!
Btw here I noticed that you'd also have to modify the `finetune/adapter_v2.py` script with so that it includes the `dtype` in the function call

rasbt (changes requested):

rasbt:
Awesome! There were few minor things with the cache and the `generate` function call in the `generate/adapter_v2.py` script, but it seems to work for me now!

Besides updating the `finetune/adapter_v2.py` script with the dtype, it seems good to go. Let me know if you want to do the fix or if I should take care of it. Happy to help.

lantiga:
Great! @Diormiu we'll get this merged as soon as the fix gets, in. If you don't have time we can push this through no problem.

rasbt:
Hey, I just needed to use the adapter_v2 script for something else and thought I'd just go ahead and implement the fix. I hope you don't mind @Diormiu. And big thanks again for the PR!!

PS: @lantiga once the tests pass this should be good to merge

rasbt (commented):
made Linear8bitLt import optional

generate/adapter_v2.py
```diff
@@ -84,10 +84,17 @@ def main(
...
+        idx=encoded,
+        max_seq_length=max_new_tokens,
+        max_new_tokens=max_new_tokens,
+        temperature=temperature,
+        top_k=top_k,
+        eos_id=tokenizer.eos_id
+    )
     t = time.perf_counter() - t0
 
-    model.reset_cache()
```
carmocca, line 90: Why is this removed?

generate/adapter_v2.py
```diff
@@ -84,10 +84,17 @@ def main(
     prompt_length = encoded.size(0)
 
     t0 = time.perf_counter()
-    y = generate(model, encoded, max_new_tokens, temperature=temperature, top_k=top_k, eos_id=tokenizer.eos_id)
+    y = generate(
+        model,
+        idx=encoded,
+        max_seq_length=max_new_tokens,
```
carmocca, line 90: What's the reasoning behind this?

rasbt: Using named arguments for easier debugging I guess

generate/adapter_v2.py
rasbt: good question

rasbt: Btw I didn't remove it. And it works fine without it. Is it necessary? @awaelchli do you know if this is essential for keeping memory low in case we have multiple generations?

carmocca: If the batch size (unimplemented) or `max_seq_length` changed between generations, then this would need to be called. Since none of these change right now, it could be removed. Or it might be good to keep for completeness

generate/adapter_v2.py
carmocca: I mean passing `max_seq_length=max_new_tokens`. This would limit it a lot as by default it will equal to the `block_size`

rasbt: I don't know, to be honest. I adopted this from the regular LLaMA adapter in the [adapter.py](https://github.com/Lightning-AI/lit-llama/blob/main/generate/adapter.py#L85) script when I originally implemented adapter_v2.py. It's also like this in `generate/lora.py`

I'd say it's okay to leave this for this PR, but then we maybe want to open an issue/PR to revisit this for ALL generate scripts?

generate/adapter_v2.py
rasbt: I added it back, with a comment, just to avoid potential bugs in case we want to play with the `max_seq_len` some day and forget this.

generate/adapter_v2.py
carmocca: The links you shared are doing it as I'm suggesting, to be clear, this is what I mean

```suggestion
```

lit_llama/adapter_v2.py
```diff
@@ -26,20 +26,34 @@ def adapter_v2_state_from_state_dict(state_dict: dict) -> dict:
 
 
 def adapter_v2_new_forward(self, input: Tensor) -> Tensor:
+    weight = self.weight
+
+    try:
+        from lit_llama.quantization import Linear8bitLt
+        if isinstance(self, Linear8bitLt):
+            weight = self.dequantize(input.dtype)
+    except:
```
carmocca, line 35: What kind of exception are we expecting here? It would be useful to list them

lit_llama/adapter_v2.py
```diff
@@ -26,20 +26,34 @@ def adapter_v2_state_from_state_dict(state_dict: dict) -> dict:
...
+        None
```
carmocca, line 36: It's more common to `pass`

```suggestion
        pass
```

lit_llama/adapter_v2.py
```diff
@@ -26,20 +26,34 @@ def adapter_v2_state_from_state_dict(state_dict: dict) -> dict:
...
 
-def adapter_v2_linear_with_bias_and_scale(layer):
-    layer.adapter_bias = torch.nn.Parameter(torch.zeros(layer.weight.shape[0]), requires_grad=True)
-    layer.adapter_scale = torch.nn.Parameter(torch.ones(layer.weight.shape[0]), requires_grad=True)
+def adapter_v2_linear_with_bias_and_scale(layer, dtype=None, quantize=False):
+    weight = layer.weight
+
+    if dtype is not None and quantize:
+        from lit_llama.quantization import Linear8bitLt
+        if isinstance(layer, Linear8bitLt):
```
carmocca, line 47: If the snippet above uses a try-catch, wouldn't you want it here too?

lit_llama/adapter_v2.py
rasbt: I don't know to be honest. I added this in case there are issues with the bitsandbytes library based on https://github.com/Lightning-AI/lit-llama/blob/main/lit_llama/quantization.py#L25

lit_llama/adapter_v2.py
rasbt: @carmocca Good point. And I just remember now why I didn't do it. I had some issues here with that.

So, some people may not be able to install `bitsandbytes`, and that shouldn't prevent people from using the adapter method without quantization. So, that's why added the `quantize` argument here. But if someone using the quantization flag, which sets `quantize=True` here AND `bitsandbytes` can not be imported. It SHOULD fail, because otherwise it would run without quantization which is not what's intended when someone uses `--quantize`. 

Now, in this case above where I used the try-except, I failed making it work with the `quantize` method because I am overriding the default forward method, and I don't think it's easily possible to add that as an argument. I am actually not sure about that and would need some help here. 

I think we actually want to remove the try-except above somehow as this is stupid and expensive if it has to fail to import something every time a forward call happens. Any ideas?

Screenshot 2023-06-07 at 4 24 53 PM

lit_llama/adapter_v2.py
carmocca: Oh that's just so you can use the project without installing bitsandbytes

rasbt: Right, I think that should then be also true when using `adapter_v2.py`. I.e., without that modification, one would have to install bitsandbytes even if the person does not want to quantize.

lit_llama/adapter_v2.py
carmocca: I see (nice image!).

We could do this with a `partial`: `partial(adapter_v2_new_forward, quantize=quantize)`

PR #324, opened 2023-05-25
Update weight download to 700B token checkpoint for the OpenLLaMA 7B

RubenFricke:


carmocca (approved):
Keep these coming! :muscle:

PR #326, opened 2023-05-26
add incremental_save to utils, use in convert_hf_checkpoint

t-vi:
This makes convert hf checkpoint slower (for 30M to bfloat16 from 2:30 (two and a half minutes) to 3:45 on my machine, but max RSS size goes from 71.9GB to 1.6GB., allowing people with less RAM to convert checkpoints easily without swapping.

### DIFF
--- a/lit_llama/utils.py
+++ b/lit_llama/utils.py
@@ -314,3 +314,160 @@ class lazy_load:
     def __exit__(self, exc_type, exc_val, exc_tb):
         del self.zf  # I don't think there is a way to force closing...
         self.zf = None
+
+
+class SavingProxyForStorage:
+    def __init__(self, obj, saver, protocol_version=5):
+        self.protocol_version = protocol_version
+        self.saver = saver
+        if not (isinstance(obj, torch.storage.TypedStorage) or torch.is_storage(obj)):
+            raise TypeError(f"expected storage, not {type(obj)}")
+
+        # this logic is taken from PyTorch 2.0+ torch/serialization.py
+        if isinstance(obj, torch.storage.TypedStorage):
+            # PT upstream wants to deprecate this eventually...
+            storage = obj._untyped_storage
+            storage_dtype = obj.dtype
+            storage_type_str = obj._pickle_storage_type()
+            storage_type = getattr(torch, storage_type_str)
+            storage_numel = obj._size()
+        else:
+            storage = obj
+            storage_dtype = torch.uint8
+            storage_type = normalize_storage_type(type(obj))
+            storage_numel = storage.nbytes()
+
+        storage_key = saver._write_storage_and_return_key(storage)
+        location = torch.serialization.location_tag(storage)
+
+        self.storage_info = (
+            "storage",
+            storage_type,
+            storage_key,
+            location,
+            storage_numel,
+        )
+
+    def __reduce_ex__(self, protocol_version):
+        raise RuntimeError("this should be handled with out of band")
+
+
+class SavingProxyForTensor:
+    def __init__(self, tensor, saver, protocol_version=5):
+        self.protocol_version = protocol_version
+        self.reduce_ret_fn, (storage, *other_reduce_args) = tensor.__reduce_ex__(
+            protocol_version
+        )
+        assert isinstance(
+            storage, torch.storage.TypedStorage
+        ), "Please check for updates"
+        storage_proxy = SavingProxyForStorage(
+            storage, saver, protocol_version=protocol_version
+        )
+        self.reduce_args = (storage_proxy, *other_reduce_args)
+
+    def __reduce_ex__(self, protocol_version):
+        if protocol_version != self.protocol_version:
+            raise RuntimeError(
+                f"Unexpected protocol version: expected {self.protocol_version}, got {protocol_version}"
+            )
+        return self.reduce_ret_fn, self.reduce_args
+
+
+class IncrementalPyTorchPickler(pickle.Pickler):
+    def __init__(self, saver, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.storage_dtypes = {}
+        self.saver = saver
+        self.id_map = {}
+
+    # this logic is taken from PyTorch 2.0+ torch/serialization.py
+    def persistent_id(self, obj):
+        # FIXME: the docs say that persistent_id should only return a string
+        # but torch store returns tuples. This works only in the binary protocol
+        # see
+        # https://docs.python.org/2/library/pickle.html#pickling-and-unpickling-external-objects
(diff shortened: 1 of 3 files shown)


t-vi:
@lantiga is this approximately what you had in mind?

I'm not quite sure about the history between conversion and padding but the automatic padding, but so the HF checkpoints from OpenAssistant are with 16 more tokens and one has to manually add the embedding size and provided the "padded" size to not get 64.

So far I have tried to stick to the original processing as much as possible, an obvious alternative taking advantage of the lazy loading could be to open all the files at once, merge the lazy-loaded SDs and then just iterate over the model keys to grab the stuff we want and count progress in terms of "number of weights".

lantiga (approved):
Amazing @t-vi, this is actually way cooler than what I had in mind.
I don't think we care about time here, it's a one-off thing, as much as we care about actually being able to convert successfully.

Seeing how convert is now implemented, I'm wondering if we could wrap the ExitStack magic into something that further shields users from that.

t-vi:
> Seeing how convert is now implemented, I'm wondering if we could wrap the ExitStack magic into something that further shields users from that (not for the current PR).

For context (ha): ExitStack is the recommended way of dealing with a variable number of context managers.

lit_llama/utils.py
```diff
@@ -314,3 +314,160 @@ def __enter__(self):
...
+        self.storage_info = (
+            "storage",
+            storage_type,
+            storage_key,
+            location,
+            storage_numel,
+        )
+
+    def __reduce_ex__(self, protocol_version):
+        raise RuntimeError("this should be handled with out of band")
```
carmocca, line 352: What does this message mean?

lit_llama/utils.py
```diff
@@ -314,3 +314,160 @@ def __enter__(self):
...
+class SavingProxyForStorage:
+    def __init__(self, obj, saver, protocol_version=5):
+        self.protocol_version = protocol_version
+        self.saver = saver
+        if not (isinstance(obj, torch.storage.TypedStorage) or torch.is_storage(obj)):
+            raise TypeError(f"expected storage, not {type(obj)}")
+
+        # this logic is taken from PyTorch 2.0+ torch/serialization.py
+        if isinstance(obj, torch.storage.TypedStorage):
+            # PT upstream wants to deprecate this eventually...
```
carmocca, line 328: I see that TypedStorage is already deprecated in `torch/storage.py`. Can you make sure that we aren't raising internal deprecation errors to users of lit-llama?

lit_llama/utils.py
```diff
@@ -314,3 +314,160 @@ def __enter__(self):
...
+
+class SavingProxyForTensor:
+    def __init__(self, tensor, saver, protocol_version=5):
+        self.protocol_version = protocol_version
+        self.reduce_ret_fn, (storage, *other_reduce_args) = tensor.__reduce_ex__(
+            protocol_version
+        )
+        assert isinstance(
+            storage, torch.storage.TypedStorage
+        ), "Please check for updates"
```
carmocca, line 363: If I understood correctly, currently, a storage is a `TypedStorage` which is being phased out. So you added this check because it will eventually be swapped for something else.

So when this raises, we need to remove this check and remove the `if isinstance(obj, torch.storage.TypedStorage)` code path in the other classes.

Is that right?

lit_llama/utils.py
```diff
@@ -314,3 +314,160 @@ def __enter__(self):
...
+            return ("storage", storage_type, storage_key, location, storage_numel)
+
+        return None
+
+
+class incremental_save:
+    def __init__(self, name):
+        self.name = name
+        self.zipfile = torch._C.PyTorchFileWriter(str(name))
+        self.have_saved = False
```
carmocca, line 438: nit: has_saved?

lit_llama/utils.py
```diff
@@ -314,3 +314,160 @@ def __enter__(self):
...
+        data_value = data_buf.getvalue()
+        self.zipfile.write_record("data.pkl", data_value, len(data_value))
+        self.have_saved = True
+
+    def _write_storage_and_return_key(self, storage):
+        if self.have_saved:
+            raise RuntimeError("have already saved")
+        key = self.next_key
+        self.next_key += 1
+        name = f"data/{key}"
```
carmocca, line 465: above (L457) you save with the .pkl suffix. Should this have it too for consistency?

tests/test_utils.py
```diff
@@ -47,6 +47,25 @@ def test_lazy_load_subclass(lit_llama):
                 torch.testing.assert_close(actual._load_tensor(), expected)
 
 
+def test_incremental_write(lit_llama):
+    import lit_llama.utils
+
+    sd = {str(k): torch.randn(5, 10) for k in range(3)}
+    sd_expected = {k: v.clone() for k, v in sd.items()}
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        path = pathlib.Path(tmpdirname)
```
carmocca, line 56: These 2 lines can be replaced with the `tmp_path` pytest fixture

carmocca (approved):
Really cool

carmocca (approved):
Really cool

lit_llama/utils.py
t-vi: No! It's not a pkl. (This is from the .pt format, what you see if you run `unzip -l checkpoint.pt`.)

lit_llama/utils.py
t-vi: Yes.

lit_llama/utils.py
t-vi: Probably could be an assert,  reduce_ex should not be called because the storage is not pickled but saved as a out of band member in the ZIP file.

tests/test_utils.py
t-vi: Thank you!

lit_llama/utils.py
t-vi: I think we are not seeing any because we are not using `Tensor.storage()` (or otherwise creating TypedStorages).
Do you see any deprecation messages (in the test or conversion)?

lit_llama/utils.py
t-vi: Haha, indeed I should not pretend object members are from an inner monologue of the objects code. Thank you for saving me from this.

t-vi:
So I updated the PR with the help of your comments, if I missed anything, I'd appreciate a hint.

lit_llama/utils.py
t-vi: :sweat_smile:

carmocca:
Thank you @t-vi!

t-vi:
Thank you for review and merge, @lantiga @carmocca

PR #336, opened 2023-05-29
Fix typo

Andrei-Aksionov:
Well, this is a lil bit embarrassing. 
I don't know English that well so for quite a while I've been relying on `cSpell` extension for VSCode. Recently I've noticed that not all typos are highlighted. The reason was in the option `"cSpell.allowCompoundWords": true`. Now of course I get much more highlights, but I think it worth it.

So I decided to revisit `lora.py` file and to fix typos.
In addition changed `finetuning` -> `fine-tuning` 'cause based on google search no one uses word `finetuning` 😄 .
Plus word `inferencing` has not the meaning that I expected so I also changed it.


lantiga (approved):
Thanks, it sure looks better after the pass :-)

PR #343, opened 2023-05-30
`incremental_save` fixes

carmocca:
See comments

### DIFF
--- a/lit_llama/utils.py
+++ b/lit_llama/utils.py
@@ -1,10 +1,10 @@
 """Utility functions for training and inference."""
 
 import functools
-from pathlib import Path
 import pickle
 import warnings
 from io import BytesIO
+from pathlib import Path
 
 import torch
 import torch.utils._device
@@ -12,7 +12,7 @@ from lightning.fabric.strategies import DeepSpeedStrategy, FSDPStrategy
 from torch.distributed.fsdp import FullStateDictConfig
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 from torch.distributed.fsdp import StateDictType
-
+from torch.serialization import normalize_storage_type
 
 llama_model_sizes = {
     4096: "7B",  # 7B n_embd=4096
@@ -327,13 +327,11 @@ class SavingProxyForStorage:
         if isinstance(obj, torch.storage.TypedStorage):
             # PT upstream wants to deprecate this eventually...
             storage = obj._untyped_storage
-            storage_dtype = obj.dtype
             storage_type_str = obj._pickle_storage_type()
             storage_type = getattr(torch, storage_type_str)
             storage_numel = obj._size()
         else:
             storage = obj
-            storage_dtype = torch.uint8
             storage_type = normalize_storage_type(type(obj))
             storage_numel = storage.nbytes()
 
--- a/scripts/convert_hf_checkpoint.py
+++ b/scripts/convert_hf_checkpoint.py
@@ -1,6 +1,7 @@
 import collections
 import contextlib
 import gc
+import json
 import shutil
 import sys
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


scripts/convert_hf_checkpoint.py
```diff
@@ -83,66 +78,57 @@ def permute(w):
...
-                    sd_key = weight_map[name]
-                    sd[sd_key] = param._load_tensor().to(dtype)
-                if not skip:
-                    sd[sd_key] = output_file.store_early(sd[sd_key])
-
-            gc.collect()
-
-        output_file.save(sd)
-
-    assert len(unprocessed_weights) == 0, f"unexpected partial weights {unprocessed_weights.key()}"
```
carmocca, line 145: replaced `.key()` which doesn't exist

scripts/convert_hf_checkpoint.py
```diff
@@ -83,66 +78,57 @@ def permute(w):
...
+                                .to(dtype)
+                            )
+                            sd[sd_key] = w
+                            del unprocessed_weights[sd_key]
+                            skip = False
+                        else:
+                            sd[sd_key] = param._load_tensor().to(dtype)
                     else:
+                        sd_key = weight_map[name]
                         sd[sd_key] = param._load_tensor().to(dtype)
```
carmocca, line 125: This else in the previous code is now removed, it should fall under the L121 else in the new code

scripts/convert_hf_checkpoint.py
```diff
@@ -50,12 +48,9 @@ def convert_hf_checkpoint(
...
 
-    # Load the json file containing weight mapping
-    pytorch_bin_map_json_path = checkpoint_dir / "pytorch_model.bin.index.json"
-    with open(pytorch_bin_map_json_path) as json_map:
-        bin_index = json.load(json_map)
-
-    bin_files = set(el for el in bin_index["weight_map"].values())
+    bin_files = list(checkpoint_dir.glob("*.bin"))
+    if not bin_files:
+        raise ValueError(f"Expected {str(checkpoint_dir)!r} to contain .bin files")
```
carmocca, line 53: Simplifying, we do this in lit-parrot

scripts/convert_hf_checkpoint.py
```diff
@@ -83,66 +78,57 @@ def permute(w):
 
     print(f"Saving to disk at {output_dir}")
     unprocessed_weights = collections.defaultdict(dict)
-    with contextlib.ExitStack() as stack:
```
carmocca, line 86: No ExitStack

lit_llama/utils.py
```diff
@@ -327,13 +327,11 @@ def __init__(self, obj, saver, protocol_version=5):
...
             # PT upstream wants to deprecate this eventually...
             storage = obj._untyped_storage
-            storage_dtype = obj.dtype
             storage_type_str = obj._pickle_storage_type()
             storage_type = getattr(torch, storage_type_str)
             storage_numel = obj._size()
         else:
             storage = obj
-            storage_dtype = torch.uint8
             storage_type = normalize_storage_type(type(obj))
```
carmocca, line 335: The import for this was missing, so clearly this code path is not used.

scripts/convert_hf_checkpoint.py
t-vi: oops. Thank you.

scripts/convert_hf_checkpoint.py
t-vi: Seems OK per se, but I wonder if there might be instances where users have other bin files for no good reason (e.g. an older version of their model had a split into smaller files).

lit_llama/utils.py
t-vi: Yes, and removing the unused `storage_dtype` is great but I would expect a future update of Pytorch to hit the second part. (but `normalize_storage_type` needs to be qualified with `torch.serialization`.... .ehem.)

t-vi (approved):
Some comments but looks great. Thank you! (Not that my review counts. :) )

awaelchli (approved):
Looks good, if good with @t-vi

scripts/convert_hf_checkpoint.py
carmocca: So far I haven't seen any hf-hub checkpoint that has that.

carmocca: That's a good point. I'll revert it

PR #344, opened 2023-05-30
add guide on finetuning any text dataset

lun-4:
Add a howto file as suggested in https://github.com/Lightning-AI/lit-llama/pull/278#issuecomment-1567322178

### DIFF
--- a/howto/unstructured_dataset.md
+++ b/howto/unstructured_dataset.md
@@ -1,0 +1,18 @@
+# Finetuning on an unstructured dataset
+
+While most scripts were made to finetune on instruction datasets, it is possible to finetune on any dataset, this is useful for experimentation while not being as expensive as training a full model.
+
+This guide is only to prepare the finetuning, as either LoRA or Adapter-v1 methods support this dataset type!
+
+## Preparation
+
+1. Gather your text into an input file named `input.txt`
+2. Prepare it into train and validation sets with the follwoing script.
+
+    ```bash
+    python scripts/prepare_any_text.py
+    ```
+
+3. Modify relevant scripts for your finetuning method under `finetune/` and `evaluate/`, setting the `instruction_tuning` variable to `False`
+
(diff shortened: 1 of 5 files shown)


howto/unstructured_dataset.md
```diff
@@ -0,0 +1,18 @@
+# Finetuning on an unstructured dataset
```
awaelchli, line 1: Looks great! We could reference this guide in the other finetune documents too. For example like this:

```
See also: [Finetuning on an unstructured dataset](../unstructured_dataset.md)
```

howto/unstructured_dataset.md
```diff
@@ -0,0 +1,18 @@
...
+
+While most scripts were made to finetune on instruction datasets, it is possible to finetune on any dataset, this is useful for experimentation while not being as expensive as training a full model.
```
awaelchli, line 3:
```suggestion
While most scripts were made to finetune on instruction datasets, it is possible to finetune on any dataset. This is useful for experimentation while not being as expensive as training a full model.
```

awaelchli (approved):
Nice :tada:

howto/unstructured_dataset.md
```diff
@@ -0,0 +1,18 @@
...
+
+This guide is only to prepare the finetuning, as either LoRA or Adapter-v1 methods support this dataset type!
+
+## Preparation
+
+1. Gather your text into an input file named `input.txt`
+2. Prepare it into train and validation sets with the follwoing script.
```
rasbt, line 10: Small typo: follwoing -> following.

It's a matter of preference, but instead of

> Prepare it into train and validation sets with the follwoing script.

you could also say

> Divide the the data into training and validation sets using the following script:

rasbt (commented):
Just a small suggestion:

lantiga (approved):
Thank you @lun-4 !

PR #348, opened 2023-05-31
Fix the multigate adapter fix for old and new checkpoint files

rasbt:
I think this fix https://github.com/Lightning-AI/lit-llama/pull/297#issuecomment-1555976464 made the new multigating in the Adapter compatible with old checkpoint file, but it then broke the new checkpoint files via this dicussion here: https://github.com/Lightning-AI/lit-llama/commit/67c1e95176928ca149ebfcb016269a401fd3ba0b#commitcomment-115733970.

This PR hopefully fixes it for both new and old checkpoint files. What do you think @awaelchli ?

### DIFF
--- a/lit_llama/adapter.py
+++ b/lit_llama/adapter.py
@@ -119,7 +119,8 @@ class CausalSelfAttention(nn.Module):
             # in case we are loading with `utils.lazy_load()`
             tensor = tensor._load_tensor() if hasattr(tensor, "_load_tensor") else tensor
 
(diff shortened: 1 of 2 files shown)


awaelchli (approved):
Of course, when I edited your PR I only tested with my old checkpoint 🤦

rasbt:
Btw if you still have that old checkpoint, can you test it there before merging? I only have a "new" checkpoint 😅 (the fix definitely works with the new checkpoint)

awaelchli:
Yes, already tested that as part of the review here. Still works 👍

PR #352, opened 2023-06-01
Adapter annotated

Andrei-Aksionov:
Hi there 👋 

This time I've managed to add annotations to, from my point of view, confusing parts of the `LLaMa-Adapter`.
Plus I've noticed that calculation of q, k, v in a single pass instead of separately multiplying input with weight matrices (x@wq, x@wk, x@wv) might confuse people so I added explanation to it also.

Didn't include arguments description (like I did it for LoRA) to be: a) more concise b) names of arguments are self-explanatory. 

The whole code as for me is much easier to understand than code for LoRa, so as a result much less annotations were needed.

---
The goal of maintainers for the repo is to keep the code as much "hackable" as possible (if I'm not mistaken). And right amount of annotations does exactly this: not to much redundant information, yet in confusing places there is a help info. So there is less possibility to get stuck and more - to use the code as a basis for further experiments.

### DIFF
--- a/lit_llama/adapter.py
+++ b/lit_llama/adapter.py
@@ -2,6 +2,9 @@
 
 LLaMA-Adapter: Efficient Fine-tuning of Language Models with Zero-init Attention
 https://arxiv.org/abs/2303.16199
+
+# TODO: put a scheme
+# TODO: put a short description
 """
 # mypy: ignore-errors
 from dataclasses import dataclass
@@ -38,6 +41,7 @@ class CausalSelfAttention(nn.Module):
             # adapter embedding layer
             self.adapter_wte = nn.Embedding(config.adapter_prompt_length, config.n_embd)
             # gate for adaption
+            # TODO: short description of why do we need gating
             self.gating_factor = torch.nn.Parameter(torch.zeros(1, config.n_head, 1, 1))
 
         self.n_head = config.n_head
@@ -57,8 +61,10 @@ class CausalSelfAttention(nn.Module):
         kv_cache: Optional[KVCache] = None,
         adapter_kv_cache: Optional[KVCache] = None,
     ) -> Tuple[torch.Tensor, Optional[KVCache], Optional[KVCache]]:
+        # TODO: put abbreviations meaning
         B, T, C = x.size()  # batch size, sequence length, embedding dimensionality (n_embd)
 
+        # TODO: put shapes
         # calculate query, key, values for all heads in batch and move head forward to be the batch dim
         q, k, v = self.c_attn(x).split(self.n_embd, dim=2)
 
@@ -67,6 +73,8 @@ class CausalSelfAttention(nn.Module):
         q = q.view(B, T, self.n_head, head_size)
         v = v.view(B, T, self.n_head, head_size)
 
+        # TODO: tell that we need to apply RoPE for every single layer
+        # TODO: put shapes
         q = apply_rope(q, rope)
         k = apply_rope(k, rope)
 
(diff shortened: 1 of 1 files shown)


lantiga (approved):
Great job @Andrei-Aksionov, everything looks clear to me! Really appreciated.

@rasbt for an extra check?

lit_llama/adapter.py
```diff
@@ -57,57 +93,81 @@ def forward(
...
 
+        # "Adapters are applied to the topmost layers to better tune the language
+        # representations with higher-level semantics".
         if self.block_idx >= self.adapter_start_layer:
             if adapter_kv_cache is not None:
-                ak, av = adapter_kv_cache
+                ak, av = adapter_kv_cache # 2 * (B, nh, aT, hs)
             else:
                 prefix = self.adapter_wte.weight.reshape(1, self.adapter_prompt_length, self.n_embd)
                 aT = prefix.size(1)
```
carmocca, line 154: We can replace `self.n_embd` calls with `C`

```suggestion
                aT = self.adapter_prompt_length
                prefix = self.adapter_wte.weight.reshape(1, aT, C)
```

lit_llama/adapter.py
```diff
@@ -57,57 +93,81 @@ def forward(
...
-                _, ak, av = self.c_attn(prefix).split(self.n_embd, dim=2)
-                ak = ak.view(1, aT, self.n_head, head_size).repeat(B, 1, 1, 1).transpose(1, 2)
-                av = av.view(1, aT, self.n_head, head_size).repeat(B, 1, 1, 1).transpose(1, 2)
+                _, ak, av = self.c_attn(prefix).split(self.n_embd, dim=2) # (1, aT, C) --> (1, aT, 3 * C) --> 3 * (1, aT, C)
```
carmocca, line 155:
```suggestion
                _, ak, av = self.c_attn(prefix).split(C, dim=2) # (1, aT, 3 * C) --> 3 * (1, aT, C)
```

lit_llama/adapter.py
Andrei-Aksionov: yeah, it's a neat fix. But, I'm not sure that it's a part of `adapter annotated` pull request.

lit_llama/adapter.py
Andrei-Aksionov: I disagree with change `self.n_embd` --> `C` as for PR related to annotations I would like to keep the code unchanged. It should be done in a separate PR I think. 

But I agree with the shorten version of shapes: to keep only output shapes. The same is true for linear projection of qkv:
https://github.com/Lightning-AI/lit-llama/blob/b3cca2c08cc5401314135d23719aa9ebaa3e64ec/lit_llama/adapter.py#L108

lit_llama/adapter.py
```diff
@@ -235,9 +295,9 @@ def forward(
                     x, rope, mask, max_seq_length, input_pos, self.kv_caches[i], self.adapter_kv_caches[i]
                 )
 
-        x = self.transformer.ln_f(x)
+        x = self.transformer.ln_f(x) # (B, T, n_embd)
 
-        logits = self.lm_head(x)  # (b, t, vocab_size)
+        logits = self.lm_head(x)  # (B, T, vocab_size)
```
carmocca, line 300: I think there's a bug in that the lm_head definition and wte should use the `padded_vocab_size` as is done for the base model

lit_llama/adapter.py
carmocca: Ah, totally okay with doing this in a separate PR. Would you like to open it or should I?

I'm using your annotations whilst I work on adapter for Lit-Parrot, and I'm finding them useful. Just wanted to share some fixes I found along the way.

lit_llama/adapter.py
Andrei-Aksionov: Again, I agree. But this annotation is correct for the current state of the code. So I propose to keep it as is for now and create a separate PR with this fix and the ones above.
I can do this right after this PR is merged.

Andrei-Aksionov: While we are discussing `lm_head` what about [weight tying](https://paperswithcode.com/method/weight-tying) like in [nanoGPT code](https://github.com/karpathy/nanoGPT/blob/master/model.py#L145)?

I assume you don't use it because in the [LLaMa repo](https://github.com/facebookresearch/llama/blob/main/llama/model.py#L205-L214) it's not implemented and thus might be a problem with loading pretrained weights. But maybe there is another reason for it. 

Asking just of curiosity 🤓 .

lit_llama/adapter.py
```diff
@@ -2,6 +2,41 @@
...
+
+                                                                             |              Prefix cross-attention
+                                                                             |
+  ┌─────────────────┐                                                        |               ┌──────────────────┐
+  ┆        x        ┆                                                        |               ┆      prefix      ┆
+  └─────────────────┘                                                        |               └──────────────────┘
+           |                                                                 |                        |
+           ▼                                                                 |                        ▼
+  ┌──────────────────┐                                                       |              ┌─────────────────────┐
+  ┆  self-attention  ┆ --------------------------------------------------------------┐      ┆  linear projection  ┆
```
rasbt, line 14: This looks great! And the write-up is also great! Thanks a lot for doing this!

rasbt (approved):

lit_llama/adapter.py
Andrei-Aksionov: So pleased to hear it. Thanks a lot! 😄

lantiga: lol

lantiga:
Alright, all good to merge!

PR #356, opened 2023-06-02
Adapter small fix

Andrei-Aksionov:
Hi there 👋 

As @carmocca mentioned in [PR #352](https://github.com/Lightning-AI/lit-llama/pull/352) some code changes need to be done:
1. Change `self.n_embd` --> `C` since this value is extracted from the shape of input variable `x` right in the beginning of the forward method.
2. Prettify reshaping of prefix.
3. This one is a biggie: `vocab_size` --> `padded_vocab_size` to align it with `lit_llama/model.py`. I assume after this checkpoints won't go south since this is just an expansion in size for better performance (I believe up to 25%). With shrinkage it would be a whole another story.

### DIFF
--- a/lit_llama/adapter.py
+++ b/lit_llama/adapter.py
@@ -147,10 +147,10 @@ class CausalSelfAttention(nn.Module):
         # "Adapters are applied to the topmost layers to better tune the language
         # representations with higher-level semantics".
         if self.block_idx >= self.adapter_start_layer:
+            aT = self.adapter_prompt_length
             if adapter_kv_cache is not None:
                 ak, av = adapter_kv_cache # 2 * (B, nh, aT, hs)
             else:
-                aT = self.adapter_prompt_length
                 prefix = self.adapter_wte.weight.reshape(1, aT, C)
                 _, ak, av = self.c_attn(prefix).split(C, dim=2) # (1, aT, 3 * C) --> 3 * (1, aT, C)
                 ak = ak.view(1, aT, self.n_head, head_size).repeat(B, 1, 1, 1).transpose(1, 2) # (B, nh, aT, hs)
@@ -159,7 +159,7 @@ class CausalSelfAttention(nn.Module):
 
             # Apply cross-attention with `query`, `adapter_key`, `adapter_value` and sum the output with the output
             # obtained from self-attention step. This is mathematically equivalent to concatenation of prefix and input as per paper.
-            amask = torch.ones(q.shape[-2], ak.shape[-2], dtype=torch.bool, device=x.device) # (T, aT)
+            amask = torch.ones(T, aT, dtype=torch.bool, device=x.device)
             # ↓ (B, nh, T, hs) @ (B, nh, aT, hs).mT --> (B, nh, T, aT) @ (B, nh, aT, hs) --> (B, nh, T, hs)
             ay = F.scaled_dot_product_attention(q, ak, av, attn_mask=amask, dropout_p=0.0, is_causal=False) # (B, nh, T, hs)
             y = y + self.gating_factor * ay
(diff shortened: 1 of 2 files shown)


lit_llama/adapter.py
```diff
@@ -150,9 +150,9 @@ def forward(
             if adapter_kv_cache is not None:
                 ak, av = adapter_kv_cache # 2 * (B, nh, aT, hs)
             else:
-                prefix = self.adapter_wte.weight.reshape(1, self.adapter_prompt_length, self.n_embd)
-                aT = prefix.size(1)
-                _, ak, av = self.c_attn(prefix).split(self.n_embd, dim=2) # (1, aT, 3 * C) --> 3 * (1, aT, C)
+                aT = self.adapter_prompt_length
```
carmocca, line 153: If you pull this out of the `if`, then in L162 you can do `torch.ones(T, aT, ...)` which would look nicer than accessing the shapes

carmocca (approved):
Thanks for this!!

lit_llama/adapter.py
Andrei-Aksionov: Thanks for the suggestion 👍 
Totally agree. 
Plus it helps with line 151. Every time I looked at the shapes in this line in my head I heard "referenced before assignment" 😄 .

Andrei-Aksionov:
Despite seeing all the green lights for merging don't do it just now. 
Tomorrow I want to check how weights are copied to be extra sure.
Right now I don't feel confident that all weights for adapter will be copied without issues and that the model will behave as expected.

awaelchli:
Thanks. Before we land this, I'd like to run the finetuning to make sure it is still training as expected. I'll do that in the next day or so.

Andrei-Aksionov:
I don't have a GPU (yeah, I know 😄 ) so I want to excuse myself in advance for any stupid questions/suggestions. Basically the problem that I wasn't able to test my suspicions with the checkpoints for this repo.

---
1. Everything should work fine simply because, as you can see from the `open_llama_7b_700bt_preview` [config](https://huggingface.co/openlm-research/open_llama_7b_700bt_preview/blob/main/config.json), the `vocab_size` is 32k which is a multiple of 64 (32k/64=500). 

2. But of course if `vocab_size` != `padded_vocab_size` then loading of pretrained weights should fail:
https://github.com/Lightning-AI/lit-llama/blob/99695716396eed07245367348a5382e73fad8834/finetune/adapter.py#L94
`load_state_dict` will not try to fill first `n` elements out of `m` (`n` < `m` where `n` - size of pretrained weights, `m` - new size).
What do I mean:
a) for embeddings if old size was 100 and now we pad the size up to 128 then we can simply fill first 100 rows in the embedding table and it will be fine 'cause this number (100) is defined by tokenizer and thus elements after max tokenizer index will not be used. So the remaining 28 elements might be initialized with any numbers.
b) almost the same is true for the `lm_head`: the only thing remaining weights (28) are needed to be initialized with zeros: logits for these non-existing tokens will be 0, after soft-max probabilites will be also 0 and thus all these 28 tokens during sampling won't be used.
But, big but, `load_state_dict` doesn't  do this as I can see. With pretrained weights it's fine, but if someone trained model from scratch and then such changes are introduced then old checkpoints are useless. 

I feel like you have already knew/discussed this, nevertheless I wanted to mention it.

---
By the way: padding up to nearest multiple of 64 in my opinion is useful only for `lm_head`. With embeddings it's basically and indexing so I don't see how we can gain performance from it.
In `nanoGPT` repo it was done for both embeddings and lm_head because of [`weight tying`](https://paperswithcode.com/method/weight-tying) --> weights are shared --> the same shape is needed during init process.

Andrei-Aksionov:
Hello @awaelchli 

> Before we land this, I'd like to run the finetuning to make sure it is still training as expected.

Any luck with this?

PR #357, opened 2023-06-02
Add Loss Averaging and no_backward_sync() in Gradient Accumulation Logic

waitzkin:
This fixes the gradient accumulation part in finetune directory.

I think averaging the loss is needed, based on the definition of gradient accumulation. 

Also, for efficiency in fine-tuning, I added no_backward_sync() where it wasn't.

### DIFF
--- a/finetune/lora.py
+++ b/finetune/lora.py
@@ -111,7 +111,7 @@ def train(
         with fabric.no_backward_sync(model, enabled=((iter_num + 1) % gradient_accumulation_iters != 0)):
             logits = model(input_ids)
             loss = loss_fn(logits, targets)
-            fabric.backward(loss / gradient_accumulation_iters )
+            fabric.backward(loss / gradient_accumulation_iters)
 
         if (iter_num + 1) % gradient_accumulation_iters == 0:
             optimizer.step()
--- a/finetune/lora.py
+++ b/finetune/lora.py
@@ -108,9 +108,10 @@ def train(
         t0 = time.time()
 
         input_ids, targets = get_batch(fabric, train_data)
-        logits = model(input_ids)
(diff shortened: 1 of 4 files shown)


finetune/lora.py
```diff
@@ -108,9 +108,10 @@ def train(
         t0 = time.time()
 
         input_ids, targets = get_batch(fabric, train_data)
-        logits = model(input_ids)
-        loss = loss_fn(logits, targets)
-        fabric.backward(loss)
+        with fabric.no_backward_sync(model, enabled=((iter_num + 1) % gradient_accumulation_iters != 0)):
+            logits = model(input_ids)
+            loss = loss_fn(logits, targets)
+            fabric.backward(loss / gradient_accumulation_iters )
```
carmocca, line 114:
```suggestion
            fabric.backward(loss / gradient_accumulation_iters)
```

carmocca (approved):
Great catch!

cc @awaelchli

awaelchli (approved):
Thanks. I remember having fixed moving the model forward under the ctx manager, but somehow the wrong version has multiplied itself into other scripts anyway 😄

finetune/lora.py
```diff
@@ -108,9 +108,10 @@ def train(
         t0 = time.time()
 
         input_ids, targets = get_batch(fabric, train_data)
-        logits = model(input_ids)
-        loss = loss_fn(logits, targets)
-        fabric.backward(loss)
+        with fabric.no_backward_sync(model, enabled=((iter_num + 1) % gradient_accumulation_iters != 0)):
```
awaelchli, line 111: FWY, I didn't add it here because this script was meant to be running on a single GPU. Anyway, doesn't hurt to have it.

PR #360, opened 2023-06-03
Integrate `Fabric.init_module(empty_weights=True)` into scripts

awaelchli:
Advertizes the use of the `init_module` context manager with empty weight initialization. This is a draft for now and requires https://github.com/Lightning-AI/lightning/pull/17627. I created this in advance to validate the PR in Lightning and to showcase the usage. 

We can decide to keep the EmptyInitOnDevice in utils for posterity and backward compatibility, but its functionality is now absorbed into `Fabric.init_module()`.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -12,7 +12,7 @@ wd = Path(__file__).parent.parent.resolve()
 sys.path.append(str(wd))
 
 from lit_llama import LLaMA, Tokenizer
-from lit_llama.utils import EmptyInitOnDevice, lazy_load, llama_model_lookup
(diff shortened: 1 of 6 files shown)


rasbt:
That's actually pretty cool. Maybe something I can highlight in the CVPR talk when loading a very large ViT

awaelchli:
@rasbt Thank you! I'm working on the usage guide/documentation in Fabric right now, I'll point you to it when its done 😃

carmocca (approved):
> We can decide to keep the EmptyInitOnDevice in utils for posterity and backward compatibility

I don't see a need for keeping it

PR #362, opened 2023-06-03
Sparse GPT Algorithm

Motsepe-Jr:
This is the sparseGPT code based on IST-DASLab project.

I followed the same coding principles as used in the lit-llama gptq code.

I created a file called sparsification which is the algorithm for SparseGPT and a folder called sparsify/sparsegpt.py to run the algorithm on the model in the checkpoint_path.

This is my first contribution to the project, If I missed some household admin I apologize in advance.

Key Notes:

    1. The source code of SparseGPT consist of a quantization algorithm similar to GPTQ, however, I removed this code because we already have GPTQ in the lit-llama source code.

2. I'm still on the waiting list for the Llama weights 7B.

### DIFF
--- a/sparsify/sparsegpt.py
+++ b/sparsify/sparsegpt.py
@@ -1,0 +1,255 @@
+# This adapts SparseGPT process: https://github.com/IST-DASLab/sparsegpt
+# E. Frantar et al SparseGPT: Massive Language Models Can Be Accurately Pruned in One-Shot, https://arxiv.org/abs/2301.00774
+# portions copyright by the authors licensed under the Apache License 2.0
+
+
+import gc
+import sys
+import time
+from pathlib import Path
+from typing import Optional
+
+import torch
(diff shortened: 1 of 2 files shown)


sparsify/sparsegpt.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+    for i, block in enumerate(model.transformer.h):
+
+        block.to(working_device)
+
+        for name in submodules_to_process:
+            print(i, name, end=" ")
+            t0 = time.perf_counter()
+            print("collecting stats", end=" ")
```
abodacs, line 80:
```python
print("collecting stats", end=" ", flush=True)
```
you can use the `flush` flag instead of `sys.stdout.flush()`

Motsepe-Jr: Okay cool, Thanks

PR #364, opened 2023-06-05
Fix Filenames on README.md

waitzkin:
This fixes some filenames in README.md. 


awaelchli (approved):
> I would appreciate any comments about this :)

I have no comments apart from:
Thank you!  🥳

PR #370, opened 2023-06-08
update weight download for open LLaMA 1T token

aniketmaurya:
1T is [here](https://huggingface.co/openlm-research/open_llama_7b).

I tried the download and conversion script with the new weights and it worked out of the box 🔥.


carmocca (approved):
Nice!

PR #387, opened 2023-06-13
Add lora batch prediction script

DavidFarago:
Add `lora_predictions.py` which
* ready a json or pytorch file with examples
* performs generation (via `lora.py`'s `main` method)
* parses the generation to separate the input from the actual completion.

Modify `lora.py`
* so its `main` method returns the output, so `main` can be reused in other scripts
* output processing according to prompt
* hyperparameters and directories.


DavidFarago:
Sorry, this PR was meant to merge into a fork.

PR #388, opened 2023-06-13
Add domain specific evaluation of predictions

DavidFarago:
Add `evaluation.py`, which performs an evaluation of a (fine-tuned) model based on our domain specific expected behavior. For this, treat a prediction as a kind of multiclass multioutput classification of its chat history. A prediction is parsed into a set of calls, each having a sequence of arguments.

Since `evaluation.py` is very different from the perplexity based evaluations in the `evaluate` folder, and expects a dataset that can be created with other scripts in this folder, put `evaluation.py` in this folder, too.

Add `README.md` to describe this prediction-evaluation-pipeline.

TODO:
1) add cosine similarity comparisons where similar text (e.g. the msg argument) is sufficient.
2) construct a full confusion matrix and report it to tensorboard.

### DIFF
--- a/generate/README.md
+++ b/generate/README.md
@@ -1,0 +1,37 @@
+# Generation with LoRA finetuned model
+
+## Single generation
+
+For a single generation (i.e. multiple inferences until a full answer has been generated)
+with our lora finetuned model, use `lora.py`.
+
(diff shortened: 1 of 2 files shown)


DavidFarago:
Sorry, this PR meant to merge into a fork.

PR #389, opened 2023-06-13
Small changes to reduce peak memory.

robieta:
LLaMA 7B is very close to the limit for sharding across 4x40GB cards, and recently several library changes have put it over the limit and it now OOMs when run. (Notably, it turns out we were cheating and [using a bit less memory than the correct implementation](https://github.com/Lightning-AI/lightning/pull/17670).) This PR introduces two small changes which prevent Shakespeare pretraining from OOMing and greatly improve its performance:

1) `foreach=False` Some PyTorch optimizers have the ability to group updates, which generally improves performance by reducing the number of kernel launches and consequently the amount of host latency. However this grouping has an increased peak memory footprint. Instead of allocating and freeing a size `X` Tensor `k` times, you allocate and free a `k * X` sized Tensor once.

2) `limit_all_gathers=True` Breaking up the optimizer step will bring the logical memory down below the OOM threshold, but if we look at memory statistics we still see reserved memory maxing out. The reason is FSDP is overzealous in launching all gathers in an attempt to overlap as much communication and compute as possible. The trouble is all of those concurrent in-flight requests take up memory; more specifically the CUDACachingAllocator caches per stream, so it's not straightforward for it to reclaim memory from the communication stream and use it in the compute stream. By restricting the number of in-flight gathers we get about a 3x performance improvement. (6-8 seconds / step -> 1.5-3 seconds / step)

### DIFF
--- a/pretrain/shakespeare.py
+++ b/pretrain/shakespeare.py
@@ -47,7 +47,7 @@ block_size = 1024
 
 def main() -> None:
     auto_wrap_policy = partial(transformer_auto_wrap_policy, transformer_layer_cls={Block})
-    strategy = FSDPStrategy(auto_wrap_policy=auto_wrap_policy, activation_checkpointing=Block)
+    strategy = FSDPStrategy(auto_wrap_policy=auto_wrap_policy, activation_checkpointing=Block, limit_all_gathers=True)
 
     fabric = L.Fabric(accelerator="cuda", devices=4, precision="bf16-mixed", strategy=strategy)
     fabric.launch()
@@ -70,7 +70,7 @@ def main() -> None:
 
     model = fabric.setup_module(model)
 
-    optimizer = torch.optim.AdamW(model.parameters(), lr=learning_rate, weight_decay=weight_decay, betas=(beta1, beta2))
+    optimizer = torch.optim.AdamW(model.parameters(), lr=learning_rate, weight_decay=weight_decay, betas=(beta1, beta2), foreach=False)
     optimizer = fabric.setup_optimizers(optimizer)
 
     train(fabric, model, optimizer, train_data, val_data)
--- a/finetune/full.py
+++ b/finetune/full.py
@@ -55,7 +55,7 @@ def main(
 ):
 
     auto_wrap_policy = partial(transformer_auto_wrap_policy, transformer_layer_cls={Block})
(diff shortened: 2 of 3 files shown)


rasbt:
This is actually really nice @robieta ! Peak memory is one of the big issues in practice when finetuning LLaMA 65B / Falcon 40B as well.

Once this is merged, we should also apply this to lit-parrot!

pretrain/shakespeare.py
```diff
@@ -70,7 +70,7 @@ def main() -> None:
 
     model = fabric.setup_module(model)
 
-    optimizer = torch.optim.AdamW(model.parameters(), lr=learning_rate, weight_decay=weight_decay, betas=(beta1, beta2))
+    optimizer = torch.optim.AdamW(model.parameters(), lr=learning_rate, weight_decay=weight_decay, betas=(beta1, beta2), foreach=False)
```
carmocca, line 73: We have several other uses of optimizers in other files. Should they be updated too?

From the PR description, is the simple rule that if you are limited by memory, you want to set `foreach=False`?

Do you think it would be a good idea to come up with some heuristic for this based on the model size? In Parrot we support sizes from 70m to 40b so setting this could be detrimental to the performance of smaller models.

pretrain/shakespeare.py
```diff
@@ -47,7 +47,7 @@
 
 def main() -> None:
     auto_wrap_policy = partial(transformer_auto_wrap_policy, transformer_layer_cls={Block})
-    strategy = FSDPStrategy(auto_wrap_policy=auto_wrap_policy, activation_checkpointing=Block)
+    strategy = FSDPStrategy(auto_wrap_policy=auto_wrap_policy, activation_checkpointing=Block, limit_all_gathers=True)
```
carmocca, line 50: There are a few other files using `FSDPStrategy` in this repo, can you update them too?

awaelchli (approved):

carmocca (approved):
I can confirm it about `foreach` from experiments in lit-gpt, haven't tested `limit_all_gathers` yet.

carmocca:
Whoa I just ran those and I'm seeing a huge speedup (95% !) by setting `limit_all_gathers` for adapter too. So we should totally do it.

carmocca:
The finetune scripts in this repo are using DeepSpeed, unlike in lit-gpt where I switched it for FSDP: https://github.com/Lightning-AI/lit-gpt/pull/118. So we can merge this.

PR #398, opened 2023-06-19
add group_by_length optional feature

nuance1979:
Notes:
- I left `validate()` intact; could use `get_dataloader()` as well and get rid of `get_batch()`;
- Dependency on `transformers` can be eliminated by copying just the `LengthGroupedSampler`.


finetune/lora.py
```diff
@@ -202,6 +207,46 @@ def pad_right(x, pad_id):
...
+        return input_ids, labels
+
+
+def get_dataloader(
+    fabric: L.Fabric,
+    data: torch.Tensor,
+    micro_batch_size: int,
+    group_by_length: bool,
+):
+    from transformers.trainer_pt_utils import LengthGroupedSampler
```
lantiga, line 229: Let's copy just this class with proper attribution in the docstring (it's a very small class and transformers is a big dependency)

finetune/lora.py
```diff
@@ -202,6 +207,46 @@ def pad_right(x, pad_id):
     return x, y
 
 
+class LlamaDataset(Dataset):
```
lantiga, line 210: I'd avoid calling this LLaMADataset as it sounds a bit misleading to me.

```suggestion
class InstructionDataset(Dataset):
```

finetune/lora.py
```diff
@@ -202,6 +207,46 @@ def pad_right(x, pad_id):
...
+):
+    from transformers.trainer_pt_utils import LengthGroupedSampler
+
+    def collate_fn(batch):
+        x, y = zip(*batch)
+        batch_x = pad_sequence(x, batch_first=True)
+        batch_y = pad_sequence(y, batch_first=True, padding_value=-1)
+        return batch_x, batch_y
+
+    dataset = LlamaDataset(data)
```
lantiga, line 237:
```suggestion
    dataset = InstructionDataset(data)
```

requirements.txt
```diff
@@ -7,3 +7,4 @@ jsonargparse[signatures]  # generate.py, convert_checkpoint.py CLI
 bitsandbytes  # quantization.py
 datasets  # evaluate.py
 zstandard  # prepare_redpajama.py
+transformers
```
lantiga, line 10: Remove

lantiga (approved):
Looks good, thank you @nuance1979!
Let's do the same for the other scripts in `finetune`

Issue #150, opened 2023-04-17
Convert lit-llama weights to huggingface


DuarteMRAlves:
Hello, I was wandering whether you are planning on releasing a script to convert weights trained with this repository to the huggingface format?

Currently, huggingface is the best way to share models across the community and I think it would be very beneficial for the adoption of this framework to be able to convert from models trained with this code to huggingface.

lantiga:
Hey @DuarteMRAlves I don't disagree. It should be fairly doable to take the current conversion script and rearrange the state dict. Help welcome :-)

timothylimyl:
piling on, I think this will be useful cause hf weights can split to multi-gpu during inference which is useful for bigger models.

carmocca:
@timothylimyl Lit-Parrot supports this via FSDP, added in https://github.com/Lightning-AI/lit-parrot/commit/248d691f06d68c7e92d3230260eda0055f7dc163. Support for this could be easily ported to Lit-LlaMA

timothylimyl:
That's awesome, any plans on supporting FSDP inference for lit-llama too?

I will give it a look too to see whether will I be able to replicate what you did on lit-parrot. However, my initial intuition was that it is not that straight forward? At least my guesses would be that you need to embed some sort of heuristic to know at which layer to separate the model given the number of gpus provided.

carmocca:
Yes, but it would be better if you or somebody else from the community works on the port.

The sharding is configured via the `auto_wrap_policy` function used in the commit I linked ([PyTorch docs](https://pytorch.org/tutorials/intermediate/FSDP_tutorial.html#id2))

timothylimyl:
> Yes, but it would be better if you or somebody else from the community works on the port.
> 
> The sharding is configured via the `auto_wrap_policy` function used in the commit I linked ([PyTorch docs](https://pytorch.org/tutorials/intermediate/FSDP_tutorial.html#id2))

Any particular reason for this?

I will give it a shot when available, I am now using another repo just because I can load using hugging face auto device mapping (but reckon lit-llama is still the best cause other repo's multi-gpu training is pretty broken)

RDouglasSharp:
[...]

PR #435, opened 2023-08-05
Convert unsharded model to huggingface format

wjurayj:
This script will convert a checkpoint fine-tuned in Lit-LLaMA into a set of files that you can load & use within huggingface.

### DIFF
--- a/scripts/convert_checkpoint_hf.py
+++ b/scripts/convert_checkpoint_hf.py
@@ -1,0 +1,233 @@
+# Copyright 2022 EleutherAI and The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+import gc
+import json
+import os
+import shutil
+import warnings
+
+import torch
+
+from transformers import LlamaConfig, LlamaForCausalLM, LlamaTokenizer
+
+
+try:
+    from transformers import LlamaTokenizerFast
+except ImportError as e:
+    warnings.warn(e)
+    warnings.warn(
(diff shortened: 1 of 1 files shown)


wjurayj:
[Issue](https://github.com/Lightning-AI/lit-llama/issues/150)

wjurayj:
> Does the current code take into consideration adapter_v2 parameters like "lm_head.adapter_bias"?

No, this only works to convert the base model into a huggingface format. For a LoRA, you can just use the Lit-LLaMA script to convert it into a base model. Whereas for a different adapter, my understanding is that it should be converted to a separate huggingface module, [like here](https://huggingface.co/docs/hub/adapter-transformers).

LamOne1:
> No, this only works to convert the base model into a huggingface format. For a LoRA, you can just use the Lit-LLaMA script to convert it into a base model. Whereas for a different adapter, my understanding is that it should be converted to a separate huggingface module, [like here](https://huggingface.co/docs/hub/adapter-transformers).

is there a way to combine adapter weights with the base model then convert the whole model to `AutoModels` in transformers?

wjurayj:
> > No, this only works to convert the base model into a huggingface format. For a LoRA, you can just use the Lit-LLaMA script to convert it into a base model. Whereas for a different adapter, my understanding is that it should be converted to a separate huggingface module, [like here](https://huggingface.co/docs/hub/adapter-transformers).
> 
> is there a way to combine adapter weights with the base model then convert the whole model to `AutoModels` in transformers?

This is possible when using the LoRA adapter, with the `convert_lora_weights` script and the instructions [here](https://github.com/Lightning-AI/lit-llama/blob/main/howto/convert_lora_weights.md). However, will be less straightforward when using a prefix approach like the LLaMA-adapter paper does because this changes the architecture of the model when you run inference on it.

As it stands, my script is only equipped to convert a model fine-tuned with the full weights or using a LoRA, since these will ultimately result in a model with the same architecture as the original LLaMA model. I'm sure that it's possible to write a similar script generalizes to prefix-tuned adapter models, but I believe this would be beyond the scope of what this script aims to accomplish.

LamOne1:
Thank you @wjurayj for you help, I appreciate your help.

LamOne1:
Dear @wjurayj ,
I'd appreciate your help in [this issue](https://github.com/Lightning-AI/lit-llama/issues/442)

wjurayj:
@awaelchli @carmocca @lantiga How can I get this contribition added to the codebase?

AChineseBoy:
FileNotFoundError: [Errno 2] No such file or directory: 'scripts/checkpoints/open-llama/tokenizer.model'

Why is there always an error? I have downloaded the tokenizer.model and placed it in the 7B folder.

Issue #438, opened 2023-08-07
This codebase has so many errors it is completely useless and unusable


Abecid:
precision = "bf16-true"

is unsupported but in the code for lora-finetuning

fabric.init_module

causes a does not exist error.

with fabric.device 

results in an error in full fine-tuning script

Overall very poor experience and poor documentation. Garbage

aniketmaurya:
This is a very constructive feedback. Kudos 👏

rasbt:
I don't know about the `with fabric.device` but let me address the other two

1. `precision = "bf16-true"`

2. `fabric.init_module`

with more explicit warnings and suggestions via a PR shortly.

rasbt:
@Abecid What error are you getting with bfloat16. I think it's only supported in Ampere and newer, but it appears that it now also works on older T4's and CPU. Just tested it. Maybe it's a PyTorch version thing. 

If you have time and don't mind spending a few more minutes, could you let me know the error code you are getting and PyTorch version to look into it further? I could then add a more explicit warning to save the hassle for future users.

PR #444, opened 2023-08-08
Check torch and lightning versions before running models

rasbt:
I think that a lot of weird errors with uninformative error messages that users encounter (e.g., see #438) are due to package mismatches. Since we are using the latest features from torch and Fabric, code kind of runs but then there's this weird error here and there, or even a missing feature.

So, I propose only allowing to run the scripts if the minimal versions are installed. If a user doesn't have these required PyTorch and Lightning versions, there is a message informing the user about that.

### DIFF
--- a/datasets
+++ b/datasets
@@ -1,1 +1,0 @@
-/data/shared/datasets/
--- a/datasets
+++ b/datasets
@@ -1,0 +1,1 @@
+/data/shared/datasets/
--- a/lit_llama/utils.py
+++ b/lit_llama/utils.py
@@ -520,23 +520,7 @@ def get_packages(pkgs):
 
 def check_python_packages():
 
-    d = {
-        'torch': '2.0.0',
-        'lightning': '2.1.0.dev0',
-    }
-
-    versions = get_packages(d.keys())
-
-    wrong_package_triggered = False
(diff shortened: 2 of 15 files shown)


datasets
```diff
@@ -0,0 +1 @@
+/data/shared/datasets/
```
Borda, line 1: is this needed?

Borda (approved):

datasets
rasbt: no must have been an accident

lit_llama/utils.py
```diff
@@ -494,3 +495,48 @@ def _write_storage_and_return_key(self, storage):
...
+        if actual_ver == 'N/A':
+            continue
+        actual_ver, suggested_ver = version_parse(actual_ver), version_parse(suggested_ver)
+        if actual_ver < suggested_ver:
+            print(f'[FAIL] {pkg_name} {actual_ver}, please upgrade to {suggested_ver}')
+            wrong_package_triggered = True
+        else:
+            print(f'[OK] {pkg_name} {actual_ver}')
+
+    if wrong_package_triggered:
```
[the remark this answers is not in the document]
rasbt: Super nice simplification 😅

PR #453, opened 2023-09-10
fix requirements + aggregate setup & req. & config in pyproject

Borda:
- fixed requirements/version so it can be run (needed after the latest Torch release)
- update to standard Python project configuration and as an advantage it allows us to add optional settings for pytest/back/etc. without adding a new file

### DIFF
--- a/.github/azure-gpu-tests.yml
+++ b/.github/azure-gpu-tests.yml
@@ -50,7 +50,7 @@ jobs:
         python -c "import torch ; mgpu = torch.cuda.device_count() ; assert mgpu == 2, f'GPU: {mgpu}'"
       displayName: 'Image info & NVIDIA'
 
-    - script: pip install .[dev] pytest
+    - script: pip install .[all] pytest
       displayName: 'Install dependencies'
 
     - bash: pytest -v --durations=10 --disable-pytest-warnings --strict-markers --color=yes
--- a/.github/azure-gpu-tests.yml
+++ b/.github/azure-gpu-tests.yml
@@ -50,7 +50,7 @@ jobs:
         python -c "import torch ; mgpu = torch.cuda.device_count() ; assert mgpu == 2, f'GPU: {mgpu}'"
       displayName: 'Image info & NVIDIA'
 
-    - script: pip install .[all] pytest
+    - script: pip install ".[all]" "pytest"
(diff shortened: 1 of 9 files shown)


.github/azure-gpu-tests.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         python -c "import torch ; mgpu = torch.cuda.device_count() ; assert mgpu == 2, f'GPU: {mgpu}'"
       displayName: 'Image info & NVIDIA'
 
-    - script: pip install pytest -r requirements.txt
+    - script: pip install .[all] pytest
```
rasbt, line 53: Does this need to be `pip install ".[all]" pytest` (I guess it depends on the shell)

pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+[project]
+name = "lit-llama"
+version = "0.1.0"
+description = "Implementation of the LLaMA language model"
+license = {text = "Apache-2.0"}
+authors = [
+    { name = "Lightning AI", email = "community@lightning.ai" }
+]
+readme = "README.md"
+requires-python = ">=3.8"
```
rasbt, line 14: In LitGPT we require >= 3.10 because we used some features not supported in 3.9 (I don't remember which ones); so maybe also make this 3.10 here to be on the safe side

rasbt (approved):

.github/azure-gpu-tests.yml
Borda: Linux not but Mac yes... anyway we can have it :)
```suggestion
    - script: pip install ".[all]" "pytest"
```

pyproject.toml
Borda:
```suggestion
requires-python = ">=3.10"
```

pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+    "bitsandbytes",
+]
+classifiers = [
+    "Topic :: Text Processing"
+]
+
+[project.optional-dependencies]
+all = [
+    "tqdm",  # convert_checkpoint.py
+    "numpy",  # train.py dataset memmap
```
Borda, line 28:
```suggestion
    "numpy <2.0",  # train.py dataset memmap
```

PR #476, opened 2023-12-16
Fix: Variable name

dhdbsrlw:
Hello.

I think variable's name should be changed in 'lit_llama/utils.py.'

When I implement your code, the existing code causes error. Then I debugged with printing. 

There was no key named 'transformer.wte.weight' in checkpoint.

Thank you.

### DIFF
--- a/lit_llama/utils.py
+++ b/lit_llama/utils.py
@@ -28,7 +28,7 @@ def llama_model_lookup(checkpoint: dict) -> str:
     
     Checks the width of the lm_head.weight matrix, as these uniquely identify the model.
     """
-    embedding_size = checkpoint['transformer.wte.weight'].shape[1]
+    embedding_size = checkpoint['lm_head.weight'].shape[1]
     return llama_model_sizes[embedding_size]
 
 


lit_llama/utils.py
```diff
@@ -28,7 +28,7 @@ def llama_model_lookup(checkpoint: dict) -> str:
     
     Checks the width of the lm_head.weight matrix, as these uniquely identify the model.
     """
-    embedding_size = checkpoint['transformer.wte.weight'].shape[1]
+    embedding_size = checkpoint['lm_head.weight'].shape[1]
```
rasbt, line 31:
```suggestion
    if 'transformer.wte.weight' in checkpoint:
        embedding_size = checkpoint['transformer.wte.weight'].shape[1]
    elif 'lm_head.weight' in checkpoint:
        embedding_size = checkpoint['lm_head.weight'].shape[1]
    else:
        raise ValueError("Neither 'transformer.wte.weight' nor 'lm_head.weight' found in the checkpoint")
```

rasbt (changes requested):
Can't reproduce it at the moment but perhaps we can have a compromise here that remains backwards compatible.

PR #478, opened 2024-01-05
Add license at the top of files

lantiga:
lit-llama is being used in other projects, often vendored (individual files taken out and reused).

This is not a bad thing per se, but we need to make sure the license and authorship are clear when this happens.

This PR adds licenses at the top of every file, just like we do for PyTorch Lightning and Fabric.


awaelchli (approved):

lantiga:
Ubuntu is failing on the torch.compile test, does this ring any bell?

Borda (approved):

